PR #1, opened 2020-04-15
Kiera Allen

KieraAllen:
I realized too late that my tests passed for how I wrote them, but they weren't written exactly like the interaction pattern.

### DIFF
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -10,4 +10,9 @@ class FoodTruck
   def check_stock(item)
     @inventory[item]
   end
+
+  def stock(item, amount)
+    @inventory[item] += amount
+  end
+
 end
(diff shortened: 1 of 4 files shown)


lib/food_truck.rb
```diff
@@ -0,0 +1,18 @@
+class FoodTruck
+  attr_reader :name,
+              :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
memcmahon, line 7: Great use of the default value!

lib/item.rb
```diff
@@ -0,0 +1,11 @@
+class Item
+  attr_reader :item,
+              :name,
+              :price
+
+  def initialize(item_info)
+    @item = item_info
+    @name = item_info[:name]
+    @price = item_info[:price][1..-1].to_f
```
memcmahon, line 9: Great job with this - you could also use `delete('$')` in the off chance that a price might not include a `$` (we would not trick you like this on a challenge, but something to think about)

PR #2, opened 2020-04-15
Ross Perry

perryr16:


lib/event.rb
```diff
@@ -0,0 +1,70 @@
...
+    #2 vendors, more than 50
+    overstocked_items = []
+    total_inventory.each do |item, details|
+      overstocked_items << item if details[:quantity] > 50 && details[:food_trucks].length > 1
+    end
+    overstocked_items
+  end
+
+  def sorted_item_list
+    total_inventory.map { |item, details| item.name}.sort
```
memcmahon, line 44: The logic here is great, though we generally don't see methods tagged on to a curly brace.

lib/food_truck.rb
```diff
@@ -0,0 +1,30 @@
+class FoodTruck
+
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = {}
```
memcmahon, line 6: If you set this hash up with a default value of 0, it would simplify the stock and check_stock methods below :)

lib/event.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def date
+    date = Date.today
+    date.strftime("%d/%m/%Y")
+  end
+
+  def sell(item, count)
```
memcmahon, line 52: This method looks super clean!

lib/event.rb
```diff
@@ -0,0 +1,70 @@
...
+    if total_inventory[item][:quantity] > count
+      total_inventory[item][:food_trucks].each do |truck|
+
+        items_sold = truck.inventory[item] if truck.inventory[item] < count
+        items_sold = count if truck.inventory[item] > count
+
+        truck.sell(item, items_sold)
```
memcmahon, line 59: Great use of a helper method!

lib/item.rb
```diff
@@ -0,0 +1,13 @@
+class Item
+
+  attr_reader :name
+  def initialize(details)
+    @name = details[:name]
+    @price = details[:price]
+  end
+
+  def price
+    @price.sub("$","").to_f
```
memcmahon, line 10: you could also use `@price.delete('$').to_f`

test/event_test.rb
```diff
@@ -0,0 +1,236 @@
...
+    food_truck3.stock(item1, 65)
+    event.add_food_truck(food_truck1)
+    event.add_food_truck(food_truck2)
+    event.add_food_truck(food_truck3)
+    assert_equal 148.75, food_truck1.potential_revenue
+    assert_equal 345.00, food_truck2.potential_revenue
+    assert_equal 243.75, food_truck3.potential_revenue
+  end
+
+  def test_case_name
```
memcmahon, line 112: Try not to leave this default test name :)

test/event_test.rb
```diff
@@ -0,0 +1,236 @@
...
+
+    expected = ["Apple Pie (Slice)", "Banana Nice Cream", "Peach Pie (Slice)", "Peach-Raspberry Nice Cream"]
+    assert_equal expected, event.sorted_item_list
+  end
+
+  def test_it_returns_date
+    event = Event.new("South Pearl Street Farmers Market")
+    event.stubs(:date).returns("24/02/2020")
+    assert_equal "24/02/2020", event.date
+  end
```
memcmahon, line 202: Instead of stubbing the `date` method (which really makes this a non-test), think about stubbing `Date.today` so that you can prove that your strftime is formatted correctly: 

```ruby
Date.stubs(:today).returns(Date.new(2020, 02, 24))
event = Event.new("South Pearl Street Farmers Market")
assert_equal "24/02/2020", event.date
```

PR #3, opened 2020-04-15
Alex Pariseau

arpariseau:
Completed iteration 4.

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -6,4 +6,8 @@ class Event
     @food_trucks = []
   end
 
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,9 @@
+class Event
+  attr_reader :name, :food_trucks
+
+  def initialize(name)
+    @name = name
+    @food_trucks = []
+  end
+
+end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -10,4 +10,8 @@ class Event
     @food_trucks << food_truck
   end
 
+  def food_truck_names
+    @food_trucks.map {|truck| truck.name}
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -14,4 +14,8 @@ class Event
     @food_trucks.map {|truck| truck.name}
   end
 
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all {|truck| truck.inventory.include?(item)}
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -30,10 +30,18 @@ class Event
 
   def get_items
     all_items = []
-    @food_trucks.each do |truck|
-      all_items << truck.inventory.keys
-    end
+    @food_trucks.each {|truck| all_items << truck.inventory.keys}
     all_items.flatten.uniq
+  end
+
+  def overstocked_items
+    overstocked = []
+    total_inventory.each do |item, item_qty|
+      if item_qty[:quantity] > 50 && item_qty[:food_trucks].length > 1
+        overstocked << item
+      end
+    end
+    overstocked
   end
 
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -51,4 +51,18 @@ class Event
     get_items.map {|item| item.name}.sort
   end
 
+  def sell(item, quantity)
+    return false if !get_items.include?(item)
+    return false if total_inventory[item][:quantity] < quantity
+    @food_trucks.each do |truck|
+      if quantity > truck.check_stock(item)
+        quantity -= truck.check_stock(item)
+        truck.stock(item, (0 - truck.check_stock(item)))
+      else
+        truck.stock(item, (0 - quantity))
+      end
+    end
+    true
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -18,4 +18,22 @@ class Event
     @food_trucks.find_all {|truck| truck.inventory.include?(item)}
   end
 
+  def total_inventory
+    inventory = {}
+    get_items.each do |item|
+      item_quantity = @food_trucks.sum {|truck| truck.check_stock(item)}
+      trucks = food_trucks_that_sell(item)
+      inventory[item] = {quantity: item_quantity, food_trucks: trucks}
+    end
+    inventory
+  end
+
+  def get_items
+    all_items = []
+    @food_trucks.each do |truck|
+      all_items << truck.inventory.keys
+    end
+    all_items.flatten.uniq
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -57,9 +57,9 @@ class Event
     @food_trucks.each do |truck|
       if quantity > truck.check_stock(item)
         quantity -= truck.check_stock(item)
-        truck.stock(item, (0 - truck.check_stock(item)))
+        truck.stock(item, -truck.check_stock(item))
       else
-        truck.stock(item, (0 - quantity))
+        truck.stock(item, -quantity)
       end
     end
     true
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,9 +1,12 @@
+require 'date'
+
 class Event
-  attr_reader :name, :food_trucks
+  attr_reader :name, :food_trucks, :date
 
   def initialize(name)
     @name = name
     @food_trucks = []
+    @date = Date.today.strftime("%d/%m/%Y")
   end
 
   def add_food_truck(food_truck)
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -44,4 +44,8 @@ class Event
     overstocked
   end
 
+  def sorted_item_list
+    get_items.map {|item| item.name}.sort
+  end
+
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -14,4 +14,10 @@ class FoodTruck
     @inventory[item]
   end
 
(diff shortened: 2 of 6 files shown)


timomitchel:
I like seeing you committing throughout and using TDD. Your commit messages do not follow convention though. They should be written in the imperative mood. This is an article that outlines the why and gives some great advice on how to write good commit messages: https://www.freecodecamp.org/news/writing-good-commit-messages-a-practical-guide/

lib/event.rb
```diff
@@ -0,0 +1,68 @@
+require 'date'
+
+class Event
+  attr_reader :name, :food_trucks, :date
+
+  def initialize(name)
+    @name = name
+    @food_trucks = []
+    @date = Date.today.strftime("%d/%m/%Y")
```
timomitchel, line 9: Nice job converting this Date object into the properly formatted string.

lib/event.rb
```diff
@@ -0,0 +1,68 @@
...
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    @food_trucks.map {|truck| truck.name}
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all {|truck| truck.inventory.include?(item)}
```
timomitchel, line 21: Exactly the two tools that get the job done here. `find_all and `include?` together is a common pattern.

lib/event.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def total_inventory
+    inventory = {}
```
timomitchel, line 25: Great job building this nested hash and using the return value of `check_stock` and `food_trucks_that_sell`.

lib/event.rb
```diff
@@ -0,0 +1,68 @@
...
+      item_quantity = @food_trucks.sum {|truck| truck.check_stock(item)}
+      trucks = food_trucks_that_sell(item)
+      inventory[item] = {quantity: item_quantity, food_trucks: trucks}
+    end
+    inventory
+  end
+
+  def get_items
+    all_items = []
+    @food_trucks.each {|truck| all_items << truck.inventory.keys}
```
timomitchel, line 36: I like this helper method. Check out the `flat_map` enumerable which is a bit more performant and flattens as you iterate.

lib/event.rb
```diff
@@ -0,0 +1,68 @@
...
+    all_items.flatten.uniq
+  end
+
+  def overstocked_items
+    overstocked = []
+    total_inventory.each do |item, item_qty|
+      if item_qty[:quantity] > 50 && item_qty[:food_trucks].length > 1
```
timomitchel, line 43: Solid job pulling on the `total_inventory`, splitting the hash in the block variable, and coming up with the correct condition.

lib/event.rb
```diff
@@ -0,0 +1,68 @@
...
+    overstocked
+  end
+
+  def sorted_item_list
+    get_items.map {|item| item.name}.sort
+  end
+
+  def sell(item, quantity)
+    return false if !get_items.include?(item)
+    return false if total_inventory[item][:quantity] < quantity
```
timomitchel, line 56: Love the use of the guard clauses here.

lib/event.rb
```diff
@@ -0,0 +1,68 @@
...
+    @food_trucks.each do |truck|
+      if quantity > truck.check_stock(item)
```
timomitchel, line 58: Elegant approach here.

lib/food_truck.rb
```diff
@@ -0,0 +1,23 @@
...
+    @inventory[item] += amount
+  end
+
+  def check_stock(item)
+    @inventory[item]
+  end
+
+  def potential_revenue
+    revenue = 0
+    @inventory.each {|item, amount| revenue += item.price * amount}
```
timomitchel, line 19: This was a great use case for using the `sum` enum.

lib/food_truck.rb
```diff
@@ -0,0 +1,23 @@
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
timomitchel, line 6: Nice job creating the hash with the default here.

lib/item.rb
```diff
@@ -0,0 +1,9 @@
+class Item
+attr_reader :name, :price
+
+  def initialize(item_info)
+    @name = item_info[:name]
+    @price = item_info[:price].delete("$").to_f.round(2)
```
timomitchel, line 6: This is the preferred spot to convert to the float, so that we have a data type we can perform math on for each Item object.

test/event_test.rb
```diff
@@ -0,0 +1,113 @@
...
+  @item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+  @item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+  @item3 = Item.new({name: "Peach-Raspberry Nice Cream", price: "$5.30"})
+  @item4 = Item.new({name: "Banana Nice Cream", price: "$4.25"})
+  @food_truck1.stock(@item1, 35)
+  @food_truck1.stock(@item2, 7)
+  @food_truck2.stock(@item3, 25)
+  @food_truck2.stock(@item4, 50)
+  @food_truck3.stock(@item1, 65)
+  @food_truck3.stock(@item3, 10)
```
timomitchel, line 24: Nice job utilizing the setup method here to speed yourself up.

test/event_test.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  def test_food_truck_names
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal ["Rocky Mountain Pies", "Ba-Nom-a-Nom", "Palisade Peach Shack"], @event.food_truck_names
+  end
+
+  def test_food_trucks_that_sell
+    @event.add_food_truck(@food_truck1)
```
timomitchel, line 51: Something to consider is adding a method that runs your `add_food_truck` setup. We don't want it in setup, because it would break our initial attributes test, but just like any other ruby class, we can create a helper method and then call it in every method that needs it:
```
def add_trucks
@event.add_food_truck(@food_truck1)
@event.add_food_truck(@food_truck2)
@event.add_food_truck(@food_truck3)
end
```
Then we call add_trucks in each test method that needs that functionality.

test/event_test.rb
```diff
@@ -0,0 +1,113 @@
...
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    item1_quantity = {quantity: 100, food_trucks: [@food_truck1, @food_truck3]}
+    item2_quantity = {quantity: 7, food_trucks: [@food_truck1]}
+    item3_quantity = {quantity: 35, food_trucks: [@food_truck2, @food_truck3]}
+    item4_quantity = {quantity: 50, food_trucks: [@food_truck2]}
+    expected = {@item1 => item1_quantity, @item2 => item2_quantity,
+                @item3 => item3_quantity, @item4 => item4_quantity}
+    assert_equal expected, @event.total_inventory
```
timomitchel, line 68: Great setup and execution here.

test/event_test.rb
```diff
@@ -0,0 +1,113 @@
...
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal [@item1], @event.overstocked_items
+  end
+
+  def test_sorted_item_list
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal ["Apple Pie (Slice)", "Banana Nice Cream",
```
timomitchel, line 89: This is a good use case for using an `expected` variable.

test/event_test.rb
```diff
@@ -0,0 +1,113 @@
...
+         "Peach Pie (Slice)", "Peach-Raspberry Nice Cream"], @event.sorted_item_list
+  end
+
+  def test_date
+    Date.stubs(:today).returns(Date.new(2020, 02, 24))
```
timomitchel, line 94: Nice job stubbing this.

test/event_test.rb
```diff
@@ -0,0 +1,113 @@
...
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal false, @event.sell(@item1, 200)
+    assert_equal false, @event.sell(item5, 1)
+    assert_equal true, @event.sell(@item4, 5)
+    assert_equal 45, @food_truck2.check_stock(@item4)
+    assert_equal true, @event.sell(@item1, 40)
+    assert_equal 0, @food_truck1.check_stock(@item1)
+    assert_equal 60, @food_truck3.check_stock(@item1)
```
timomitchel, line 110: Great robust testing.

test/food_truck_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    food_truck2 = FoodTruck.new("Ba-Nom-a-Nom")
+    item3 = Item.new({name: "Peach-Raspberry Nice Cream", price: "$5.30"})
+    item4 = Item.new({name: "Banana Nice Cream", price: "$4.25"})
+    food_truck2.stock(item3, 25)
+    food_truck2.stock(item4, 50)
+
+    food_truck3 = FoodTruck.new("Palisade Peach Shack")
+    food_truck3.stock(@item1, 65)
+
+    assert_equal 148.75, @food_truck.potential_revenue
```
timomitchel, line 56: Nice job creating the objects you needed for this test.

PR #4, opened 2020-04-15
Stella Bonnie 

stellakunzang:
Completed iteration 3!

`event.sell` method in iteration 4 created an infinite loop that I wasn't able to resolve, so the last assertion in the test isn't passing.

### DIFF
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -12,4 +12,13 @@ class FoodTruck
   def stock(item, quantity)
     inventory[item] += quantity
   end
+
+  def potential_revenue
+    revenue = 0
+    inventory.each do |item, count|
+      revenue += (item.price * count)
+    end
+    revenue 
+  end
+
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -18,7 +18,10 @@ class FoodTruck
     inventory.each do |item, count|
       revenue += (item.price * count)
     end
-    revenue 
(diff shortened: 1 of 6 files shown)


lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new { |hash, key| hash[key] = 0 }
```
memcmahon, line 5: Great use of a default value here - could be simplified to `Hash.new(0)`

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
...
+    inventory[item]
+  end
+
+  def stock(item, quantity)
+    inventory[item] += quantity
+  end
+
+  def potential_revenue
+    revenue = 0
+    inventory.each do |item, count|
```
memcmahon, line 18: Great implentation of each over a hash - you could also use `sum`

lib/item.rb
```diff
@@ -0,0 +1,9 @@
+class Item
+  attr_reader :name, :price
+
+  def initialize(item_info)
+    @name = item_info[:name]
+    @price = item_info[:price].gsub(/[^\d\.]/, '').to_f
```
memcmahon, line 6: Totally works, but you could also use `.delete('$')` :)

PR #5, opened 2020-04-15
Sage Lee

sagemlee:


lib/event.rb
```diff
@@ -0,0 +1,51 @@
...
+    @food_trucks = []
+  end
+
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    @food_trucks.map do |food_truck|
+      food_truck.name
```
aziobrow, line 15: great use of `map`; perfect tool for this job

lib/event.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+  end
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |truck|
+      truck.inventory.include?(item)
+    end
```
aziobrow, line 21: also a great use for `find_all`.  i'd probably lean towards `has_key?` vs `include?` on the inventory, to be more semantic.

lib/event.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def total_inventory
+    total = {}
+    @food_trucks. each do |truck|
+      truck.inventory.each do |item|
+        total[item] = 0
+      end
+    end
+    
+    total
```
aziobrow, line 32: this is a tricky one for sure! i'd do something like this:

```
  def total_inventory
    result = {}
        
    items.each do |item|
    end
    result
  end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,34 @@
+require 'pry'
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = {}
```
aziobrow, line 6: This is good!

`@inventory = Hash.new(0)`

lib/food_truck.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def check_stock(item)
+    if @inventory[item] != nil
+        @inventory[item]
+    else
+      @inventory[item] = 0
+    end
```
aziobrow, line 14: nice! it's great that you caught that you had to check for whether or not the item existed in @inventory.

aziobrow: if you had the default value set, as I mentioned above, you don't actually have to check for the include (since it would be 0 if it didn't exist).

So this method could just be:
```
  def check_stock(item)
      @inventory[item]
 end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def stock(item, amount)
+    if @inventory[item] != nil
+      @inventory[item] += amount
+    else
+      @inventory[item] = 0
+      @inventory[item] += amount
+    end
```
aziobrow, line 23: same thing here-- this is great for the way you have it written!

```
def stock(item, price)
    @inventory[item] += price
 end 
```

lib/food_truck.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def potential_revenue
+    rev = @inventory.map do |k,v|
```
aziobrow, line 27: great implementation (although i probably wouldn't use map/sum, because then you're technically iterating twice, since `sum` is an enumerable too, when you only needed to iterate once).  this works really nice and is overall an elegant solution, if not completely optimal.

try not to take shortcuts like this though; `rev` is unclear as an abbreviation, and `k,v` is pretty meaningless (go with |item, quantity|, which is what they really are). 

This is how i'd do it: 

```
def potential_revenue
    total = 0
         item.price * quantity
     end
     total 
end 
```

notice how much easier that is to read/understand.

lib/item.rb
```diff
@@ -0,0 +1,9 @@
+class Item
+    attr_reader :name, :price
+
+  def initialize(info)
+    @name = info[:name]
+    @price = info[:price].sub("$","").to_f
```
aziobrow, line 6: great!

test/event_test.rb
```diff
@@ -0,0 +1,72 @@
...
+    assert_equal 243.75, @food_truck3.potential_revenue
+  end
+
+  def test_it_can_calculate_total_inventory
+    @food_truck3.stock(@item3, 10)
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal ({@item1 => {quantity: 100 , food_trucks: [@food_truck1, @food_truck3]}, @item2 => {quantity: 7 , food_trucks: [@food_truck1, @food_truck2]}, @item3 => {quantity: 50 , food_trucks: [@food_truck1]}, @item4 => {quantity: 35 , food_trucks: [@food_truck2]} }), @event.total_inventory
+  end
```
aziobrow, line 70: great

test/food_truck_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  def test_it_has_readable_attributes
+    assert_equal "Rocky Mountain Pies", @food_truck.name
+    assert_equal ({}), @food_truck.inventory
+  end
+
+  def test_it_can_stock_items
+    assert_equal 0, @food_truck.check_stock(@item1)
+    @food_truck.stock(@item1, 30)
+    assert_equal ({@item1 => 30}), @food_truck.inventory
+    assert_equal 30, @food_truck.check_stock(@item1)
```
aziobrow, line 26: since they are two separate methods, i'd split the test/assertions for `check_stock`  from the `inventory` test/assertions.

test/food_truck_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def test_it_can_multiple_items
```
aziobrow, line 29: then i'd just put these assertions into either the `check_stock` test or the `inventory` test-- as in, rather than writing a separate test for multiples, just test the multiples within the appropriate test that you already tested the single for.

test/item_test.rb
```diff
@@ -0,0 +1,23 @@
...
+  def test_it_exists
+    assert_instance_of Item, @item1
+    assert_instance_of Item, @item2
+  end
+
+  def test_it_has_readable_attributes
+    assert_equal "Apple Pie (Slice)", @item2.name
+    assert_equal 2.50, @item2.price
+  end
+
```
aziobrow, line 22: 👍

lib/food_truck.rb
aziobrow: i thiiink i learned that you can do sum on a hash (you didn't used to be able to so I think i've just been out of touch)

```
def potential_revenue
         item.price * quantity
     end
end 
```

sagemlee:
@aziobrow Thank you for all of the feedback! I have been just shy of iteration 3 every challenge so i'm really motivated to get over that hump.  I will definitely be studying the next few nights!

PR #6, opened 2020-04-15
Danny Ramos

muydanny:
:/

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,41 @@
+class Event
+
+  attr_reader :name,
+              :food_trucks
+
+  def initialize(name)
+    @name = name
+    @food_trucks = []
+  end
+
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    names = []
+    @food_trucks.each do |food_truck|
+      names << food_truck.name
+    end
+    names
+  end
+  # `food_trucks_that_sell` that takes an
+  # argument of an item represented as a String.
+  # It will return a list
+  # of FoodTrucks that have that item in stock.
+
+  def food_trucks_that_sell(desired_item)
+    food_truck_list = []
+    @food_trucks.each do |food_truck|
+      food_truck.inventory.each do |item|
+        if item.first.name == desired_item.name
+          food_truck_list << food_truck
+        end
+      end
+    end
+    food_truck_list
+  end
+
+
+
+end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -1,0 +1,30 @@
+class FoodTruck
+
+  attr_reader :name,
+              :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
+    @default_stock = 0
+  end
+
+  def check_stock(item)
+    @inventory.sum do |item|
+      item[1] += @default_stock
+    end
+  end
+
+  def stock(item, quantity)
+    @inventory[item] = quantity
+  end
+
+  # the sum of all their items'
+  # price * quantity.
+
+  def potential_revenue
+    @inventory.sum do |inventory|
+      inventory.first.price * inventory.last
+    end
+  end
+end
--- a/lib/item.rb
+++ b/lib/item.rb
@@ -1,0 +1,14 @@
+class Item
+
+  attr_reader :name
+
+
+  def initialize(item_info)
+    @name = item_info[:name]
+    @price = item_info[:price]
+  end
+
+  def price
+    @price.gsub(/[^\d\.]/, '').to_f
+  end
+end
--- a/test/event_test.rb
+++ b/test/event_test.rb
@@ -1,0 +1,74 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/item'
+require './lib/food_truck'
+require './lib/event'
+
+class EventTest < Minitest::Test
+  def setup
+    @event = Event.new("South Pearl Street Farmers Market")
+    @food_truck1 = FoodTruck.new("Rocky Mountain Pies")
+    @food_truck2 = FoodTruck.new("Ba-Nom-a-Nom")
+    @food_truck3 = FoodTruck.new("Palisade Peach Shack")
+    @item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    @item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+    @item3 = Item.new({name: "Peach-Raspberry Nice Cream", price: "$5.30"})
+    @item4 = Item.new({name: "Banana Nice Cream", price: "$4.25"})
+    end
+
+    def test_it_exists
+      assert_instance_of Event, @event
+    end
+
+    def test_it_has_attributes
+      assert_equal "South Pearl Street Farmers Market", @event.name
+      assert_equal [], @event.food_trucks
+    end
+
+    def test_event_has_food_truck
+      @food_truck1.stock(@item1, 35)
+      @food_truck1.stock(@item2, 7)
+      @food_truck2.stock(@item4, 50)
+      @food_truck2.stock(@item3, 25)
+      @food_truck3.stock(@item1, 65)
+      @event.add_food_truck(@food_truck1)
+      @event.add_food_truck(@food_truck2)
+      @event.add_food_truck(@food_truck3)
+      assert_equal [@food_truck1, @food_truck2, @food_truck3], @event.food_trucks
+    end
+
+    def test_event_can_give_food_truck_names
+      @event.add_food_truck(@food_truck1)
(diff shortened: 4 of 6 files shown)


timomitchel:
Would have loved to see commits throughout, and the tests committed first as the directions called for.

lib/event.rb
```diff
@@ -0,0 +1,41 @@
...
+      names << food_truck.name
+    end
+    names
+  end
+  # `food_trucks_that_sell` that takes an
+  # argument of an item represented as a String.
+  # It will return a list
+  # of FoodTrucks that have that item in stock.
+
+  def food_trucks_that_sell(desired_item)
```
timomitchel, line 27: Nice job with this method! This works, but you can save yourself some time and some lines with a combination of `find_all` and `includes?`:
```
@food_trucks.find_all {|food_truck| food_truck.inventory.include?(item)}
```

lib/event.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def food_truck_names
+    names = []
+    @food_trucks.each do |food_truck|
+      names << food_truck.name
+    end
+    names
+  end
+  # `food_trucks_that_sell` that takes an
```
timomitchel, line 22: I love this pseudo coding! It worked!

lib/food_truck.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def stock(item, quantity)
+    @inventory[item] = quantity
+  end
+
+  # the sum of all their items'
+  # price * quantity.
+
+  def potential_revenue
+    @inventory.sum do |inventory|
```
timomitchel, line 26: You should be splitting this hash by key and value pairs in the block variable here:
```
item.price * quantity
end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,30 @@
+class FoodTruck
+
+  attr_reader :name,
+              :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
```
timomitchel, line 8: We were hoping to see this hash initialized with a default value of 0 based on the interaction patter. You can do this two ways:
```
Hash.new(0)
```
```
@inventory.default = 0

File: lib/food_truck.rb [hunk 5]
```diff
@@ -0,0 +1,30 @@
...
+    @default_stock = 0
+  end
+
+  def check_stock(item)
+    @inventory.sum do |item|
```
timomitchel, line 13: This method needed to return the value for the item key in our inventory hash:
```
@inventory[item]
```

lib/food_truck.rb
```diff
@@ -0,0 +1,30 @@
...
+      item[1] += @default_stock
+    end
+  end
+
+  def stock(item, quantity)
+    @inventory[item] = quantity
```
timomitchel, line 19: We were looking for something different here:
```
if @inventory.has_key?(item)
@inventory[item] += quantity
else
@inventory[item] = quantity
end
```
When the condition returns true, we add to the current value of item, when it's false, we add the item key to the hash and assign it a quantity.

lib/item.rb
```diff
@@ -0,0 +1,14 @@
...
+  attr_reader :name
+
+
+  def initialize(item_info)
+    @name = item_info[:name]
+    @price = item_info[:price]
+  end
+
+  def price
+    @price.gsub(/[^\d\.]/, '').to_f
```
timomitchel, line 12: Oh you fancy with your regex ay?

lib/item.rb
```diff
@@ -0,0 +1,14 @@
+class Item
+
+  attr_reader :name
+
+
+  def initialize(item_info)
+    @name = item_info[:name]
+    @price = item_info[:price]
```
timomitchel, line 8: It's preferred to do the conversion for `@price` right in initialize and then use and `attr_reade`r for price:
`@price =info[:price].gsub("$", "").to_f`
This way the conversion only runs once, not every time you call the #price method.

test/event_test.rb
```diff
@@ -0,0 +1,74 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/item'
+require './lib/food_truck'
+require './lib/event'
+
+class EventTest < Minitest::Test
+  def setup
```
timomitchel, line 8: Nice job utilizing a setup method here to save time.

test/event_test.rb
```diff
@@ -0,0 +1,74 @@
...
+      assert_equal "South Pearl Street Farmers Market", @event.name
+      assert_equal [], @event.food_trucks
+    end
+
+    def test_event_has_food_truck
+      @food_truck1.stock(@item1, 35)
+      @food_truck1.stock(@item2, 7)
+      @food_truck2.stock(@item4, 50)
+      @food_truck2.stock(@item3, 25)
+      @food_truck3.stock(@item1, 65)
```
timomitchel, line 33: No need for the setup from 29 - 33 here. We just want to make sure we can add food trucks, but they don't need inventory for this test to work properly.

test/event_test.rb
```diff
@@ -0,0 +1,74 @@
...
+      @event.add_food_truck(@food_truck3)
+      assert_equal [@food_truck1, @food_truck2, @food_truck3], @event.food_trucks
+    end
+
+    def test_event_can_give_food_truck_names
+      @event.add_food_truck(@food_truck1)
+      @event.add_food_truck(@food_truck2)
+      @event.add_food_truck(@food_truck3)
+      expected = ["Rocky Mountain Pies", "Ba-Nom-a-Nom", "Palisade Peach Shack"]
+      assert_equal expected, @event.food_truck_names
```
timomitchel, line 45: Nice job here.

test/event_test.rb
```diff
@@ -0,0 +1,74 @@
...
+      @food_truck1.stock(@item1, 35)
+      @food_truck1.stock(@item2, 7)
+      @food_truck2.stock(@item4, 50)
+      @food_truck2.stock(@item3, 25)
+      @food_truck3.stock(@item1, 65)
+      @event.add_food_truck(@food_truck1)
+      @event.add_food_truck(@food_truck2)
+      @event.add_food_truck(@food_truck3)
+      assert_equal [@food_truck1, @food_truck3], @event.food_trucks_that_sell(@item1)
+      assert_equal [@food_truck2], @event.food_trucks_that_sell(@item4)
```
timomitchel, line 58: Great job robustly testing this method!

test/event_test.rb
```diff
@@ -0,0 +1,74 @@
...
+      @food_truck1.stock(@item2, 7)
+      @food_truck2.stock(@item4, 50)
+      @food_truck2.stock(@item3, 25)
+      @food_truck3.stock(@item1, 65)
+      @food_truck3.stock(@item3, 10)
+      @event.add_food_truck(@food_truck1)
+      @event.add_food_truck(@food_truck2)
+      @event.add_food_truck(@food_truck3)
+      expected = { @item1 => {quantity => 100, food_trucks => [food_truck1, food_truck3]}}
+      assert_equal expected, @event.total_inventory
```
timomitchel, line 72: The setup is incorrect here. We needed to have each of the stocked items as keys in our hash.

test/food_truck_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    def test_it_can_stock
+       @food_truck.stock(@item1, 30)
+       expected = {@item1 => 30}
+       assert_equal expected, @food_truck.inventory
+       assert_equal 30, @food_truck.check_stock(@item1)
+       @food_truck.stock(@item1, 25)
+       # assert_equal 55, @food_truck.check_stock(@item1)
+       @food_truck.stock(@item2, 12)
+       expected = {@item1 => 55, @item2 => 12}
+       # assert_equal expected, @food_truck.inventory
```
timomitchel, line 40: This should have caused more alarm. Always best to get a method working before moving on.

test/food_truck_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def test_potential_revenue_of_food_truck
+      @food_truck1.stock(@item1, 35)
+      @food_truck1.stock(@item2, 7)
+      @food_truck2.stock(@item4, 50)
+      @food_truck2.stock(@item3, 25)
+      @food_truck3.stock(@item1, 65)
+      assert_equal 148.75, @food_truck1.potential_revenue
```
timomitchel, line 49: Nice job with this test.

lib/item.rb
muydanny: lmaooo this was such a shot in the dark with a google search and when it worked I literally jumped out  my chair but yeah Im fancy

muydanny:
@timomitchel Thanks Tim! Methods took me longer than I wanted but this assessment felt immensely better than previous ones. Pry really does save lives lol Although Im not quite there yet with more work and studying Im feeling good going into the last week.

PR #7, opened 2020-04-15
Jane Greene

janegreene:


lib/food_truck.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def check_stock(item)
+    in_stock = 0
+    @inventory.each do |k, v|
+        if k.name == item.name
+          in_stock = v
+        end
+   end
+   in_stock
+  end
```
memcmahon, line 17: This definitely works, but an easier and more straightforward solution would be to use brackets to access the value at a specific key:

```ruby
if @inventory[item].nil?
  0
else
  @inventory[item]
end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,31 @@
...
+
+
+  def stock(item, count)
```
memcmahon, line 20: Great job with this method!

lib/item.rb
```diff
@@ -0,0 +1,9 @@
+class Item
+  attr_reader :name, :price
+
+  def initialize(argument)
+    @name = argument[:name]
+    @price = argument[:price].delete( "$" ).to_f
```
memcmahon, line 6: Great!

PR #8, opened 2020-04-15
Michelle :) 

foleymichelle9:


lib/food_truck.rb
```diff
@@ -0,0 +1,24 @@
+class FoodTruck
+
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
memcmahon, line 7: Great use of a default value here!  It really helps with the other methods in this class!

lib/food_truck.rb
```diff
@@ -0,0 +1,24 @@
...
+    @inventory[item]
+  end
+
+  def stock(item, amount)
+    @inventory[item] += amount
+  end
+
+  #the sum of all their items' price * quantity.
+  def potential_revenue
+    @inventory.inject(0) do |total, inventory|
```
memcmahon, line 20: Great use of `inject` - you could also use `sum` :)

lib/event.rb
```diff
@@ -0,0 +1,80 @@
...
+    @food_trucks.each do |food_truck|
+      total_food_items << food_truck.inventory.keys
+    end
+    total_food_items = total_food_items.flatten.uniq
+  end
+
+  def find_item_quantity(item)
+   x = items_and_quantity
+   x[item]
+  end
```
memcmahon, line 35: Functionality here looks good, but I don't love the variable `x` 😆

lib/event.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def total_inventory
+    event_items = {}
+   find_total_food_items.each do |item|
+     event_items[item] = {quantity: find_item_quantity(item),
+                           food_trucks: food_trucks_that_sell(item)}
+   end
+   event_items
+  end
```
memcmahon, line 64: Good use of helper methods to build this hash - you could also do something like this:

```ruby
  def total_inventory
    inventory = {}
    @food_trucks.each do |truck|
      truck.inventory.each do |item, quantity|
        inventory[item] = {quantity: 0, food_trucks: []} if inventory[item].nil?
        inventory[item][:quantity] += quantity
        inventory[item][:food_trucks] << truck
      end
    end
    inventory
  end
```

lib/event.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def overstocked_items
+    over_stocked = total_inventory.find_all do |items|
```
memcmahon, line 67: When iterating over a hash, you can use two block variables, so instead of `|items|` you could use `|item, quantity|` so you have more distinct variable names to use in the block 😄

lib/item.rb
```diff
@@ -0,0 +1,9 @@
+class Item
+
+  attr_reader :name, :price
+
+  def initialize(attributes)
+    @name = attributes[:name]
+    @price = attributes[:price][1..-1].to_f
```
memcmahon, line 7: Great!  You could also use `attributes[:price].delete('$').to_f`

PR #9, opened 2020-04-15
Derek Borski

dborski:


test/item_test.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def test_it_exists
+
+    assert_instance_of Item, @item1
+  end
+
+  def test_it_has_readable_attributes
+
+    assert_equal 'Peach Pie (Slice)', @item1.name
+    assert_equal "$3.75", @item1.price
```
memcmahon, line 21: Keep a close eye on the interaction pattern - we were looking for price to return a float `3.75` in this case.

PR #10, opened 2020-04-15
Norma Lopez

IamNorma:


test/item_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/item'
+
+class ItemTest < Minitest::Test
```
timomitchel, line 5: This testing looks good throughout the class.

lib/item.rb
```diff
@@ -0,0 +1,8 @@
+class Item
+  attr_reader :name, :price
+  
+  def initialize(item_params)
+    @name = item_params[:name]
+    @price = item_params[:price].delete("$").to_f
```
timomitchel, line 6: Nice job doing this conversion within initialize. This is preferred, so that the price is always a float.

lib/food_truck.rb
```diff
@@ -0,0 +1,19 @@
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
```
timomitchel, line 6: We were hoping to see this hash initialized with a default value of `0` based on the interaction patter. You can do this two ways: 
```
Hash.new(0)
```
Or after creating the hash, you could assign the default value by calling `.default` on the hash
```
@inventory.default = 0
```

lib/food_truck.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def check_stock(item)
```
timomitchel, line 9: This method needed to return the value for the item key in our inventory hash:
`@inventory[item]`
the above is what we were looking for this method to return.

lib/food_truck.rb
```diff
@@ -0,0 +1,19 @@
...
+    if @inventory[item] == nil
+      return 0
+    else @inventory[item]
+    end
+  end
+
+  def stock(item, amount)
+    @inventory[item] = amount
```
timomitchel, line 17: This method needed to either assign a new value as you did or add to an existing value if the item key already existed. We needed to create some kind of control flow to achieve this:
```
if @inventory[item] == 0
     @inventory[item] = amount
else
     @inventory[item] += amount
end

File: lib/food_truck.rb [hunk 6]
```diff
@@ -0,0 +1,19 @@
...
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
+  end
+
+  def check_stock(item)
+    if @inventory[item] == nil
+      return 0
+    else @inventory[item]
```
timomitchel, line 12: We never put a condition or functionality on the same line as our `else`. Else never takes a condition, but instead executes some code when the `if` conditional is false.

test/food_truck_test.rb
```diff
@@ -0,0 +1,49 @@
...
+    assert_equal "Rocky Mountain Pies", food_truck.name
+  end
+
+  def test_it_starts_with_no_inventory
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+
+    assert_equal ({}), food_truck.inventory
+  end
+
+  def test_it_can_check_stock
```
timomitchel, line 26: This is the correct setup here, but I imagine that this test was giving you some trouble on the final assertion.

test/food_truck_test.rb
```diff
@@ -0,0 +1,49 @@
...
+
+
+    assert_equal 0, food_truck.check_stock(item1)
+    food_truck.stock(item1, 30)
+    assert_equal 30, food_truck.check_stock(item1)
+    food_truck.stock(item1, 25)
+    assert_equal 55, food_truck.check_stock(item1)
+  end
+
+  def test_it_can_stock
```
timomitchel, line 38: Again, this is a solid test setup. It looks like this interaction tripped you up a bit.

timomitchel:
Another search that may have helped here: `ruby accessing key that does not exist in hash returns 0 instead of nil`

PR #11, opened 2020-04-15
Jack Gould 

jhgould:


lib/event.rb
```diff
@@ -0,0 +1,67 @@
...
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    names = []
+    @food_trucks.each do |truck|
+      names << truck.name
+    end
+    names
```
aziobrow, line 18: this works well-- `map` would be an even better tool. 

```
def food_truck_names
    @food_trucks.map {|truck| truck.name}
end 
```

lib/event.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def food_trucks_that_sell(item)
+    trucks = []
+    @food_trucks.each do |truck|
+      if truck.inventory.include?(item)
+        trucks << truck
+      end
+    end
+    trucks
```
aziobrow, line 28: also very good, and `find_all` would be the best tool here. 

```
def food_trucks_that_sell(item)
    @food_trucks.find_all do |truck|
         truck.inventory.include?(item) 
     end
end 
```

lib/event.rb
```diff
@@ -0,0 +1,67 @@
...
+    list.each do |k, v|
+      @food_trucks.each do |truck|
+        truck.inventory.each do |item|
+          if item[0].name == k
+            list[k] << truck
+          end
+        end
+      end
+    end
+    list
```
aziobrow, line 52: cool-- this one was a tricky one for sure. you have a few pieces that aren't quite hooked up correctly, but really good progress and effort. 

this is how i'd solve it:

```
  def total_inventory
    result = {}
        
    items.each do |item|
      total_quantity = @food_trucks.sum {|truck| truck.inventory[item]}
    end
    result
 end
```

lib/event.rb
```diff
@@ -0,0 +1,67 @@
...
+
+
+  def sorted_item_list
+    list = []
+    @food_trucks.each do |truck|
+      truck.inventory.each do |item|
+        list << item[0].name
+      end
+    end
+    list.uniq.sort
```
aziobrow, line 63: good work with this one. this one was a bit tricky too, and it does exactly what you need for it to do.

That's definitely the important part, but do note that you're iterating a bunch of times here-- you're using nested eachs, so you're iterating several times for every truck and then you're iterating again at the end with the sort. This is fine with a small set of data as you have here, but if this were a set of data like you're working with in Futbol, this would be really inefficient (if you had 4 trucks, you'd be iterating a total of 5 separate iterations over 5 different collections-- once for the inventory of each of the 4 trucks and once more over the list of resulting items)

Something like this would really cut it down:

```
def sorted_item_list
    @food_trucks.flat_map {|truck| truck.inventory.keys}.uniq.sort
end
```
This is just two iterations (once over the trucks, and once over the resulting items).

Hopefully this made sense-- let me know if it doesn't and I can try to clarify tomorrow.

lib/food_truck.rb
```diff
@@ -0,0 +1,22 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
aziobrow, line 5: great use of a default value here. ends up making your life much easier.

lib/food_truck.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def check_stock(item)
+    @inventory[item]
```
aziobrow, line 9: beautiful

lib/food_truck.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def stock(item, amount)
+    @inventory[item] += amount
```
aziobrow, line 13: perfect

lib/food_truck.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def potential_revenue
+    total = @inventory.sum do |item|
+      item[0].price * item[1]
+    end
+    total
+  end
```
aziobrow, line 21: this is a good implementation (i really like the use of sum), although when iterating over a hash, you should split apart the block variables into the key and value-- it makes it much easier to read/understand.
```
def potential_revenue
         item.price * quantity
     end
end 
```

see how that's much easier to figure out what's going on then trying to figure out what item[0].price/item[1] is.

lib/item.rb
```diff
@@ -0,0 +1,13 @@
+class Item
+  attr_reader :name
+
+  def initialize(data)
+    @name = data[:name]
+    @price = data[:price].delete"$"
```
aziobrow, line 6: you can just call the `.to_f` at the end here and then just use the attr_reader.

test/event_test.rb
```diff
@@ -0,0 +1,137 @@
...
+     food_truck1 = FoodTruck.new("Rocky Mountain Pies")
+     item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+     item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+     item3 = Item.new({name: "Peach-Raspberry Nice Cream", price: "$5.30"})
+     item4 = Item.new({name: "Banana Nice Cream", price: "$4.25"})
+     food_truck1.stock(item1, 35)
+     food_truck1.stock(item2, 7)
+     food_truck2 = FoodTruck.new("Ba-Nom-a-Nom")
+     food_truck2.stock(item4, 50)
+     food_truck2.stock(item3, 25)
```
aziobrow, line 34: you don't need to stock all of these trucks to test that food trucks can be added-- it doesn't hurt, but it's probably not a good use of your time.

test/event_test.rb
```diff
@@ -0,0 +1,137 @@
...
+     food_truck3 = FoodTruck.new("Palisade Peach Shack")
+     food_truck3.stock(item1, 65)
```
aziobrow, line 57: same here, no need to stock.

test/event_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    food_truck3 = FoodTruck.new("Palisade Peach Shack")
+    food_truck3.stock(item1, 65)
+    food_truck3.stock(item3, 10)
+    event.add_food_truck(food_truck1)
+    event.add_food_truck(food_truck2)
+    event.add_food_truck(food_truck3)
+      hash = {item1 => { 100 => [food_truck1, food_truck3]},
+              item2 => {7 => [food_truck2] },
+              item3 => {50 => [food_truck2]},
+              item4 => {35 => [food_truck2, food_truck3]} }
```
aziobrow, line 109: this doesn't quite match the interaction pattern-- the hashes for the values are missing keys.  i'd probably recommend copying the interaction pattern into your expectation and then just tweaking the specific things that you need to change (like the food truck objects).

test/event_test.rb
```diff
@@ -0,0 +1,137 @@
...
+     food_truck2 = FoodTruck.new("Ba-Nom-a-Nom")
+     food_truck2.stock(item4, 50)
+     food_truck2.stock(item3, 25)
+     food_truck3 = FoodTruck.new("Palisade Peach Shack")
+     food_truck3.stock(item1, 65)
+     food_truck3.stock(item3, 10)
+     event.add_food_truck(food_truck1)
+     event.add_food_truck(food_truck2)
+     event.add_food_truck(food_truck3)
+     array = ["Apple Pie (Slice)", "Banana Nice Cream", "Peach Pie (Slice)", "Peach-Raspberry Nice Cream"]
```
aziobrow, line 132: as a general rule, don't name things their data type (like `array` here). everything has a data type, so it's meaningless. something like `sorted_names` would be better here.

test/food_truck_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    assert_equal 30, food_truck.check_stock(item1)
+
+    assert_equal ({item1 => 30}), food_truck.inventory
+
+    food_truck.stock(item1, 25)
+    assert_equal 55, food_truck.check_stock(item1)
+
+    food_truck.stock(item2, 12)
+    assert_equal 12, food_truck.check_stock(item2)
+    assert_equal ({item1 => 55, item2 => 12}), food_truck.inventory
```
aziobrow, line 42: this is a good progression test to check for fully dynamic behavior. 

however, since they are two separate methods, i'd split the test/assertions for check_stock from the inventory test/assertions.

PR #13, opened 2020-04-15
Melanie Tran

melatran:
In terms of organization, I know it's probably not best practice to change up test layout midway but  food_truck3.stock(item3, 10) got me stuck so from test_total_inventory to the end does not use the test setup.

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,4 @@
+class Event
+  def initialize(name)
+  end
+end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -8,4 +8,10 @@ class Event
   def add_food_truck(food_truck_name)
     @food_trucks << food_truck_name
   end
+
+  def food_truck_names
+    @food_trucks.map do |food_truck|
+      food_truck.name
+    end
+  end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -35,9 +35,10 @@ class Event
   end
 
   def overstocked_items
-    total_inventory.select do |item, item_info|
+    overstock = total_inventory.select do |item, item_info|
       item_info[:quantity] > 50 && item_info[:food_trucks].length > 1
-    end.keys
+    end
+    overstock.keys
   end
 
   def sorted_item_list
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -14,4 +14,11 @@ class Event
       food_truck.name
     end
   end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |food_truck|
+      food_truck.check_stock(item) > 0
+    # require "pry";binding.pry
+    end
+  end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -4,4 +4,8 @@ class Event
     @name = name
     @food_trucks = []
   end
+
+  def add_food_truck(food_truck_name)
+    @food_trucks << food_truck_name
+  end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,4 +1,7 @@
 class Event
+  attr_reader :name, :food_trucks
   def initialize(name)
+    @name = name
+    @food_trucks = []
   end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -21,4 +21,33 @@ class Event
     # require "pry";binding.pry
     end
   end
+  # 
+  # def list_items
+  #   inventories = @food_trucks.map {|food_truck| food_truck.inventory}
+  #   item_names = inventories.flat_map {|inventory| inventory.keys}
+  #   item_names.uniq
+  # end
+
+  def total_inventory
+    total = {}
+    @food_trucks.each do |food_truck|
+      food_truck.inventory.each do |item, quantity|
+        total[item] = {quantity: 0, food_trucks:[]} if total[item].nil?
+        total[item][:quantity] += quantity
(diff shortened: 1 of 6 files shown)


timomitchel:
Great job committing throughout and showing us TDD!

lib/event.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def food_truck_names
+    @food_trucks.map do |food_truck|
+      food_truck.name
+    end
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |food_truck|
+      food_truck.check_stock(item) > 0
```
timomitchel, line 20: This condition works for the pattern, but a more dynamic pattern would be
```
 @food_trucks.find_all do |truck|
food_truck.inventory.include?(item)
end
```
Because a food truck may sell an item, but currently not have a quantity of it and therefore that item key exists in our inventory hash, but has a value of 0. This way we are checking that the item exists in our hash regardless of quantity.

lib/event.rb
```diff
@@ -0,0 +1,63 @@
...
+    # require "pry";binding.pry
+    end
+  end
+
+  def total_inventory
+    total = {}
```
timomitchel, line 26: Nice job getting this one running. Another implementation that does not use nested iteration:
```
def total_inventory
    # I need items as my keys, so I need to build this hash beginning with items
    #food_trucks have items - leggo
    total = {}
    items.each do |item|
      # I scope the trucks by an item with a previous method I built
      trucks_for_item = food_trucks_that_sell(item)
      # I get the total amount for each truck's inventory that includes the item
      total_amount = trucks_for_item.sum {|truck| truck.inventory[item]}
      #I build my nested hash based on the above return values for each ite,
      total[item] = {quantity: total_amount, food_trucks: trucks_for_item}
    end
    total
  end
```

lib/event.rb
```diff
@@ -0,0 +1,63 @@
...
+        total[item] = {quantity: 0, food_trucks:[]} if total[item].nil?
+        total[item][:quantity] += quantity
+        total[item][:food_trucks] << food_truck
+      end
+    end
+    total
+  end
+
+  def overstocked_items
+    overstock = total_inventory.select do |item, item_info|
```
timomitchel, line 38: Great job pulling on a previous method, splitting this hash in the block, and coming up with the correct condition.

lib/event.rb
```diff
@@ -0,0 +1,63 @@
...
+      item_info[:quantity] > 50 && item_info[:food_trucks].length > 1
+    end
+    overstock.keys
+  end
+
+  def sorted_item_list
```
timomitchel, line 44: Nice sorting.

lib/food_truck.rb
```diff
@@ -0,0 +1,21 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
timomitchel, line 5: Exactly what we were hoping to see you do here.

lib/food_truck.rb
```diff
@@ -0,0 +1,21 @@
...
+  def check_stock(item)
+    @inventory[item]
+  end
+
+  def stock(item, quantity)
+    @inventory[item] += quantity
+  end
+
+  def potential_revenue
+    @inventory.sum do |item, quantity|
```
timomitchel, line 17: Clean and effective use of sum.

lib/item.rb
```diff
@@ -0,0 +1,8 @@
+class Item
+  attr_reader :name, :price
+  
+  def initialize(item_info)
+    @name = item_info[:name]
+    @price = item_info[:price].delete("$").to_f
```
timomitchel, line 6: Nice job doing the string manipulation in initialize. This ensures the string manipulation only runs once per object and then we have a data type we can perform calculations on.

test/event_test.rb
```diff
@@ -0,0 +1,138 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/item'
+require './lib/food_truck'
+require './lib/event'
+
+class EventTest < Minitest::Test
+  def setup
```
timomitchel, line 8: Nice use of the setup method to speed things up.

test/event_test.rb
```diff
@@ -0,0 +1,138 @@
...
+
+  def test_can_get_food_trucks_that_sell_item
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal [@food_truck1, @food_truck3], @event.food_trucks_that_sell(@item1)
+  end
+
+  def test_can_get_total_inventory
+    event = Event.new("South Pearl Street Farmers Market")
```
timomitchel, line 55: I saw your comment on the PR. This is solid attention to detail. This method (and iteration 3) required a slightly different setup, so you did nothing wrong, but did exactly what I would have done when faced with that scenario.

test/food_truck_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    assert_equal ({}), @food_truck.inventory
+  end
+
+  def test_can_check_stock
+    assert_equal 0, @food_truck.check_stock(@item1)
+  end
+
+  def test_add_stock_to_inventory
+    @food_truck.stock(@item1, 30)
+    expected = {@item1 => 30}
```
timomitchel, line 30: I really like the use of the `expected` variable to help make the tests more readable.

PR #14, opened 2020-04-15
Antonio Jackson

AntonioJacksonII:


lib/event.rb
```diff
@@ -0,0 +1,40 @@
...
+                                        food_trucks: [truck]}
+          else
+            inventory_total[items[0]] = {quantity: @food_trucks.sum {|truck| truck.check_stock(items[0])},
+                                          food_trucks: [truck]}
+          end
+        end
+      end
+    end
+    inventory_total
+  end
```
memcmahon, line 39: You're definitely on the right track, but a couple of things going on here:
1. Each time you iterate over an item, you are summing the total quantity of all trucks, so any item that is sold at more than one truck will end up with more than the actual quantity.
2. You are always resetting the array of [truck] regardless of if that item already has keys set up.
3. you are iterating over items twice - once through inventory and again through items (which is actually an array of an inventory key and value)

Take a look at a possible solution below:

```ruby
    inventory_total = {}
    @food_trucks.each do |truck|
      truck.inventory.each do |items|
        if inventory_total[items[0]].nil?
          inventory_total[items[0]] = {quantity: items[1],
                                      food_trucks: food_trucks_that_sell(items[0])}
        else
          inventory_total[items[0]][:quantity] += items[1]
        end
      end
    end
    inventory_total
```

Also, when iterating over hashes, you can set up two block variables, which makes it a little easier to see what objects you have access to:

```ruby
    inventory_total = {}
    @food_trucks.each do |truck|
      truck.inventory.each do |item, quantity|
        if inventory_total[item].nil?
          inventory_total[item] = {quantity: quantity,
                                      food_trucks: food_trucks_that_sell(item)}
        else
          inventory_total[item][:quantity] += quantity
        end
      end
    end
    inventory_total
```

lib/food_truck.rb
```diff
@@ -0,0 +1,21 @@
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
memcmahon, line 6: Great use of a default value!

lib/food_truck.rb
```diff
@@ -0,0 +1,21 @@
+class FoodTruck
```
memcmahon, line 1: This class looks great!!

PR #15, opened 2020-04-15
Final Commit

GDemps:
I"m getting better outside of getting stopped on little things that take up too much of my time.  I will finish iteration 3 of the final.

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,14 @@
+class Event
+
+  attr_reader :name,
+              :food_trucks
+
+  def initialize(name)
+    @name = name
+    @food_trucks = []
+  end
+
+  def food_trucks_that_sell(item)
+
+  end
+end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -1,0 +1,28 @@
+class FoodTruck
+
+  attr_reader :name,
+              :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
+  end
+
+  def check_stock(item)
+    if @inventory[item] == nil
+      @inventory[item] = 0
+    else
+      @inventory[item]
+    end
+  end
+
+  def stock(item, price)
+    if @inventory[item] == nil
+      @inventory[item] = 0
+      @inventory[item] += price
+    else
+      @inventory[item] += price
+    end
+  end
+
+end
--- a/lib/item.rb
+++ b/lib/item.rb
@@ -1,0 +1,10 @@
+class Item
+
+  attr_reader :name,
+              :price
+
+  def initialize(item = {})
+    @name = item[:name]
+    @price = item[:price].delete_prefix("$").to_f
+  end
+end
--- a/test/test_event.rb
+++ b/test/test_event.rb
@@ -1,0 +1,19 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/food_truck'
+require './lib/item'
+require './lib/event'
+
+class FoodTruckTest < Minitest::Test
+
+  def test_instance_of_event
+    event = Event.new("South Pearl Street Farmers Market")
+    assert_instance_of Event, event
+  end
+
+  def test_name
+    event = Event.new("South Pearl Street Farmers Market")
+    assert_equal "South Pearl Street Farmers Market", event.name
+    assert_equal [], event.food_trucks
+  end
+end
--- a/test/test_foodtruck.rb
+++ b/test/test_foodtruck.rb
@@ -1,0 +1,50 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/food_truck'
+require './lib/item'
+
+class FoodTruckTest < Minitest::Test
+
+  def test_instance_of_food_truck
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    assert_instance_of FoodTruck, food_truck
+  end
+
+  def test_food_truck_name
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    assert_equal "Rocky Mountain Pies", food_truck.name
+  end
+
+  def test_food_truck_inventory
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    assert_equal ({}), food_truck.inventory
+  end
+
+  def test_food_truck_stock
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+    assert_equal "Apple Pie (Slice)", item2.name
+    assert_equal 2.50, item2.price
+  end
+
+  def test_check_stock
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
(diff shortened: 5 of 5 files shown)


lib/item.rb
```diff
@@ -0,0 +1,10 @@
+class Item
+
+  attr_reader :name,
+              :price
+
+  def initialize(item = {})
```
aziobrow, line 6: this totally works! you don't need (or really want) to set a default to a hash here--you want to require someone to give you the item_info when the item is created. Also, since this is the item class, I'd probably call the parameter `item_info` so that it's clear.  so this may seem small, but it would be preferable to do something like this:

```
 def initialize(item_info)
    @name = item_info[:name]
    @price = item_info[:price].delete_prefix("$").to_f
end
```

lib/item.rb
```diff
@@ -0,0 +1,10 @@
...
+    @name = item[:name]
+    @price = item[:price].delete_prefix("$").to_f
```
aziobrow, line 8: fantastic job with this! perfect.

lib/food_truck.rb
```diff
@@ -0,0 +1,28 @@
...
+    @name = name
+    @inventory = {}
+  end
+
+  def check_stock(item)
+    if @inventory[item] == nil
+      @inventory[item] = 0
+    else
+      @inventory[item]
+    end
```
aziobrow, line 16: nice! it's great that you caught that you had to check for whether or not the item existed in `@inventory`.

lib/food_truck.rb
```diff
@@ -0,0 +1,28 @@
+class FoodTruck
+
+  attr_reader :name,
+              :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
```
aziobrow, line 8: This is good! 

`@inventory = Hash.new(0)`

lib/food_truck.rb
aziobrow: if you had the default value set, as I mentioned above, you don't actually have to check for the nil value (since it would be 0 if it didn't exist). 

So this method could be:

```
  def check_stock(item)
      @inventory[item]
 end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+  end
+
+  def stock(item, price)
+    if @inventory[item] == nil
+      @inventory[item] = 0
+      @inventory[item] += price
+    else
+      @inventory[item] += price
+    end
```
aziobrow, line 25: same thing here-- this is great for the way you have it written!

```
def stock(item, price)
    @inventory[item] += price
 end 
```

lib/food_truck.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def check_stock(item)
+    if @inventory[item] == nil
+      @inventory[item] = 0
+    else
+      @inventory[item]
+    end
+  end
+
+  def stock(item, price)
```
aziobrow, line 19: the second argument isn't `price`-- it's actually `quantity`.  So if the key is a `tomato` item, the value is how many tomatoes this food truck has in stock. a price belongs to an item.

lib/event.rb
```diff
@@ -0,0 +1,14 @@
+class Event
+
+  attr_reader :name,
+              :food_trucks
+
+  def initialize(name)
+    @name = name
+    @food_trucks = []
```
aziobrow, line 8: 👍

test/test_event.rb
```diff
@@ -0,0 +1,19 @@
...
+require './lib/event'
+
+class FoodTruckTest < Minitest::Test
+
+  def test_instance_of_event
+    event = Event.new("South Pearl Street Farmers Market")
+    assert_instance_of Event, event
+  end
+
+  def test_name
```
aziobrow, line 14: you're testing more than name here, so this test would be better called `test_attributes`

test/test_event.rb
```diff
@@ -0,0 +1,19 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/food_truck'
+require './lib/item'
+require './lib/event'
+
+class FoodTruckTest < Minitest::Test
```
aziobrow, line 7: this is called FoodTruckTest even though this is the EventTest-- be careful. I imagine you just copied this and forgot to change the name, but you can run into issues when you have multiple classes named the same thing.

test/test_foodtruck.rb
```diff
@@ -0,0 +1,50 @@
...
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    assert_equal ({}), food_truck.inventory
+  end
+
+  def test_food_truck_stock
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+    assert_equal "Apple Pie (Slice)", item2.name
+    assert_equal 2.50, item2.price
```
aziobrow, line 28: you're not testing anything with stock here-- you're just testing that the items you created have a name and a price, which you already did in the ItemTest. So this isn't a good test in this case. 

Your test should look something like this:

```
  def test_it_can_stock_an_item
      @food_truck.stock(@item1, 30)
       assert_equal ({@item1 => 30}), @food_truck.inventory
       @food_truck.stock(@item2, 12)
        assert_equal ({@item1 => 55, @item2 => 12}), @food_truck.inventory
end 
```

test/test_foodtruck.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def test_check_stock
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    assert_equal 0, food_truck.check_stock(item1)
```
aziobrow, line 34: it's great to check that an empty inventory returns 0 as the quantity for an item that hasn't been stocked yet, but i would definitely also check an item after it's been added to ensure that the correct quantities are being retrieved.

test/test_foodtruck.rb
```diff
@@ -0,0 +1,50 @@
...
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+    assert_equal 30, food_truck.stock(item1, 30)
+    assert_equal 30, food_truck.check_stock(item1)
+    food_truck.stock(item1, 25)
+    assert_equal 55, food_truck.check_stock(item1)
+    assert_equal 12, food_truck.stock(item2, 12)
+    require "pry"
+    binding.pry
```
aziobrow, line 47: this test has the same name as the one that I commented on above-- you don't want two tests with the same name, or it will only run one.  See my comment for the one above for how this test should look.  (you should be asserting what the inventory looks like)

GDemps:
I will.  I got caught up accidentally adding an item price twice and
couldn't figure out why my answer was off.  Figured out I was adding two
prices and I will be busting my ass this weekend to time myself on all of
our previous independent challenges and practice to go for it next week.
Thanks Aurora.  The feedback helps me a bunch. Cheers!

Grant

PR #16, opened 2020-04-15
Whitney Kidd

whitneykidd:


lib/event.rb
```diff
@@ -0,0 +1,65 @@
...
+  attr_reader :name, :food_trucks
+  def initialize(name)
+    @name = name
+    @food_trucks = []
+  end
+
+  def date
+    date = Date.today
+    date.strftime("%d/%m/%Y")
+  end
```
aziobrow, line 12: great!

lib/event.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    @food_trucks.map { |food_truck| food_truck.name }
```
aziobrow, line 19: perfect tool

lib/event.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all { |food_truck| food_truck.inventory.include?(item) }
```
aziobrow, line 23: great choice of enumerable here too

lib/event.rb
```diff
@@ -0,0 +1,65 @@
...
+          inventory[items.first][:food_trucks] << food_truck
+        else
+          inventory[items.first] = {
+            quantity: items.last,
+            food_trucks: [food_truck]
+          }
+        end
+      end
+      inventory
+    end
```
aziobrow, line 40: very awesome-- this one was super tricky and you nailed it. nice use of reduce. my only recommendation would be that when you're iterating over the inventory on line 28, since that's a hash, use two block variables (|item, quantity|)-- that makes code a lot easier to read. 

As an alternative, here's my solution:

```
  def total_inventory
    result = {}
        
    items.each do |item|
      total_quantity = @food_trucks.sum {|truck| truck.inventory[item]}
    end
    result
 end
```

lib/event.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def overstocked_items
+    overstocked_items = total_inventory.find_all do |items|
+      items.last[:quantity] > 50 && items.last[:food_trucks].length > 1
+    end
+    overstocked_items.map { | items | items.first }
```
aziobrow, line 47: i'd probably use a map on line 44, use a conditional in the block, and just compact the nils out at the end-- it avoids iterating twice and gets it all done in a single iteration. 

Same here-- i'd use two block variables when iterating over a hash (lines 44 and 47)

Otherwise, fantastic work.

lib/event.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def sorted_item_list
+    total_inventory.map do |items|
+      items.first.name
+    end.sort
```
aziobrow, line 53: beautiful.

lib/event.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def sell(item, quantity)
+    if total_inventory[item][:quantity] >= quantity
+      return true
```
aziobrow, line 58: ahh, so close. remember that when you use an explicit return statement, nothing underneath it executes. so if there's enough quantity, it will immediately return true here and never execute line 59, which does the subtracting. i believe this will work if you just moved this line 58 to below line 59 (and you can just take out line 60)

lib/food_truck.rb
```diff
@@ -0,0 +1,19 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
aziobrow, line 5: great use of a default value here, makes life a lot easier.

lib/food_truck.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def check_stock(item)
+    @inventory[item]
```
aziobrow, line 9: beautiful

lib/food_truck.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def stock(item, quantity)
+    @inventory[item] += quantity
```
aziobrow, line 13: love it

lib/food_truck.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def potential_revenue
+    @inventory.sum { |item| item.first.price * item.last }
```
aziobrow, line 17: great-- again, only suggestion would be to use two block variables

```
def potential_revenue
    @inventory.sum {|item, quantity| item.price * quantity }
end 
```

see how that's much easier to figure out what's going on then trying to figure out what item.first.price/item.last is.

lib/item.rb
```diff
@@ -0,0 +1,8 @@
+class Item
+  attr_reader :name, :price
+  def initialize(attributes)
+    @name = attributes[:name]
+    @price = attributes[:price][1..-1].to_f.to_f
```
aziobrow, line 5: it works fine in this case, but i probably wouldn't use a range for removing characters from a price-- this would break if the price was more than 3 digits (like if it cost 40.75).  i'd do something like:

`attributes[:price].delete('$')`

also, this may have been a typo or something, but you don't need two `to_f`s

test/event_test.rb
```diff
@@ -0,0 +1,127 @@
...
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    sorted_list = ["Apple Pie (Slice)", "Banana Nice Cream", "Peach Pie (Slice)", "Peach-Raspberry Nice Cream"]
+    assert_equal sorted_list, @event.sorted_item_list
+  end
+
+  def test_it_can_return_date
+    @event.stubs(:date).returns("24/02/2020")
+    assert_equal "24/02/2020", @event.date
```
aziobrow, line 115: right idea, but not quite the right execution. since you stubbed `@event` to return the `date` method with the string, and then immediately assert that, you're just testing your stub. 

You'd actually want to do something like:
```
date = Date.new(2001,1,1)
`Date.stubs(:today).returns(date)`

assert_equal "01/01/2001"
```

test/event_test.rb
```diff
@@ -0,0 +1,127 @@
...
+    # today returns => assert_equal "15/04/2020", @event.date
+  end
+
+  def test_it_can_sell_items
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal false, @event.sell(@item1, 200)
+    assert_equal true, @event.sell(@item4, 5)
+    # assert_equal 45, @food_truck2.check_stock(@item4)
```
aziobrow, line 125: i'd probably want to see this test finished (including the reduced quantities) before i saw your implementation written in the class

test/food_truck_test.rb
```diff
@@ -0,0 +1,47 @@
...
+    assert_equal 25, @food_truck.check_stock(@item1)
+    inventory1 = {@item1 => 25}
+    assert_equal inventory1, @food_truck.inventory
+
+    @food_truck.stock(@item1, 30)
+    assert_equal 55, @food_truck.check_stock(@item1)
+
+    @food_truck.stock(@item2, 12)
+    inventory2 = {@item1 => 55, @item2 => 12}
+    assert_equal inventory2, @food_truck.inventory
```
aziobrow, line 38: this is a good progression test to check for fully dynamic behavior.

however, since they are two separate methods, i'd split the test/assertions for check_stock from the inventory test/assertions.

PR #17, opened 2020-04-15
Rostam Mahabadi

Rostammahabadi:


lib/event.rb
```diff
@@ -0,0 +1,51 @@
...
+    @name = name
+    @food_trucks = []
+  end
+
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    @food_trucks.each_with_object([]) do |k ,acc|
```
timomitchel, line 14: This works, but `.map` is the best enum for the job here.

lib/event.rb
```diff
@@ -0,0 +1,51 @@
...
+      acc << k.name
+    end
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all{|truck|truck.inventory.include?(item)}
```
timomitchel, line 20: This is a great combination of the best tools for the job here: `find_all` and `inclue?`

lib/event.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def sorted_item_list
+    #doesn't account for if they are in stock
+    item_names = []
+    @food_trucks.each do |food_truck|
+      food_truck.inventory.keys.each do |item|
+        if item.check_stock != 0
```
timomitchel, line 28: There was no need for this check_stock call here, which should take an argument based on the interaction pattern.

lib/event.rb
```diff
@@ -0,0 +1,51 @@
...
+        item_names << item.name
+      end
+      end
+    end
+    item_names.uniq!.sort
```
timomitchel, line 33: We would prefer `uniq` to `uniq!` here.

lib/event.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def total_inventory
+    sub_hash = {}
+    @food_trucks.each do |food_truck|
```
timomitchel, line 38: Nice job building this hash, because you have a few methods that need all of the unique items, it would have been useful to make a helper method that returns all unique items in the inventory hash for each food_truck.

lib/food_truck.rb
```diff
@@ -0,0 +1,32 @@
+class FoodTruck
+
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
```
timomitchel, line 7: We really wanted to see a default value passed to this hash upon initialization.

lib/food_truck.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  end
+
+  def check_stock(item)
+    return 0 if @inventory[item] == nil
```
timomitchel, line 12: This works, but with a default value of 0, we could simplify this entire method to `@inventory[item]`

lib/food_truck.rb
```diff
@@ -0,0 +1,32 @@
...
+    @inventory[item]
+  end
+
+  def stock(item, quantity)
+    if @inventory[item] == nil
```
timomitchel, line 17: Nice job getting this one to pass, but again, a default value simplifies this method to `@inventory[item] += quantity` which would be able to run whether the key value pair existed or not.

lib/food_truck.rb
```diff
@@ -0,0 +1,32 @@
...
+      @inventory[item] = quantity
+    else
+      @inventory[item] += quantity
+    end
+  end
+
+  def potential_revenue
+    total_revenue = 0
+    @inventory.each do |item|
```
timomitchel, line 26: This works, but this is also a great use case for the `sum` enumerable.

test/event_test.rb
```diff
@@ -0,0 +1,170 @@
...
+
+  def test_it_has_attributes
+    event = Event.new("South Pearl Street Farmers Market")
+
+    assert_equal "South Pearl Street Farmers Market", event.name
+    assert_equal [], event.food_trucks
+  end
+
+  def test_it_can_add_food_trucks
+    event = Event.new("South Pearl Street Farmers Market")
```
timomitchel, line 25: This test would benefit greatly from a `setup` method for object creation and stocking.

test/event_test.rb
```diff
@@ -0,0 +1,170 @@
...
+    food_truck2 = FoodTruck.new("Ba-Nom-a-Nom")
+    food_truck2.stock(item4, 50)
+    food_truck2.stock(item3, 25)
+    food_truck3 = FoodTruck.new("Palisade Peach Shack")
+    food_truck3.stock(item1, 65)
+    food_truck3.stock(item3, 10)
+    event.add_food_truck(food_truck1)
+    event.add_food_truck(food_truck2)
+    event.add_food_truck(food_truck3)
+    assert_equal ["Apple Pie (Slice)", "Banana Nice Cream", "Peach Pie (Slice)", "Peach-Raspberry Nice Cream"], event.sorted_item_list
```
timomitchel, line 125: This one should have been throwing a wrong number of arguments for the `check_stock` method. This worries me into thinking that you're not running your tests and following your error message.

test/food_truck_test.rb
```diff
@@ -0,0 +1,65 @@
...
+  def test_it_can_check_stock_after_stocking
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+    food_truck.stock(item1, 30)
+    assert_equal 30, food_truck.check_stock(item1)
+    food_truck.stock(item1, 25)
+    assert_equal 55, food_truck.check_stock(item1)
+    food_truck.stock(item2, 12)
+    assert_equal ({item1 => 55, item2 => 12}), food_truck.inventory
```
timomitchel, line 62: Solid testing throughout this file, but again, a setup method will speed you up.

test/event_test.rb
```diff
@@ -0,0 +1,170 @@
...
+    food_truck2.stock(item4, 50)
+    food_truck2.stock(item3, 25)
+    food_truck3 = FoodTruck.new("Palisade Peach Shack")
+    food_truck3.stock(item1, 65)
+    food_truck3.stock(item3, 10)
+    event.add_food_truck(food_truck1)
+    assert_equal ({item1=>{quantity: 35, food_trucks: [food_truck1]},item2=>{quantity:7, food_trucks:[food_truck1]}}), event.total_inventory
+    event.add_food_truck(food_truck2)
+    event.add_food_truck(food_truck3)
+    assert_equal ({item1=>{quantity: 100, food_trucks: [food_truck1,food_truck3]}, item2 =>{quantity:7, food_trucks:[food_truck1]}, item3=>{quantity:35, food_trucks:[food_truck2, food_truck3]}}), event.total_inventory
```
timomitchel, line 147: Whenever we have a data type that takes up this many lines, it's best to stack our hash and use an `expected` variable to increase readability.

timomitchel:
Nice job committing and

PR #18, opened 2020-04-15
Michael Gallup

Gallup93:
Got bogged down on iteration one for like 15 mins too long. After that, I never felt like I didn't know what to do but I ran out of time.

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,6 @@
+class Event
+  attr_reader :name
+  def initialize(name)
+    @name = name
+  end
+end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -27,4 +27,17 @@ class Event
     list.uniq.sort
   end
 
+  def total_inventory
+    totals = {}
+    sorted_item_list.each do |item|
+      totals[item] = 0
+    end
+    @food_trucks.each do |truck|
+      truck.inventory.keys.each do |key|
+        totals[key.name] += truck.inventory[key]
+      end
+    end
+
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,6 +1,12 @@
 class Event
-  attr_reader :name
+  attr_reader :name, :food_trucks
   def initialize(name)
     @name = name
+    @food_trucks = []
   end
+
+  def add_food_truck(truck)
+    @food_trucks << truck
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -19,4 +19,12 @@ class Event
     end
   end
 
+  def sorted_item_list
+    list = []
+    @food_trucks.each do |truck|
+      truck.inventory.keys.each {|key| list << key.name}
+    end
+    list.uniq.sort
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -13,4 +13,10 @@ class Event
     @food_trucks.map {|truck| truck.name}
   end
 
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |truck|
+      truck.check_stock(item) > 0
+    end
+  end
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -9,4 +9,8 @@ class Event
     @food_trucks << truck
   end
 
+  def food_truck_names
+    @food_trucks.map {|truck| truck.name}
+  end
+
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -1,0 +1,7 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = {}
+  end
+end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -22,4 +22,14 @@ class FoodTruck
       @inventory[new_item] += amount
     end
   end
+
+  def potential_revenue
+    total = 0
+    if inventory != nil
+      @inventory.each do |item|
+        total += (item[0].price * item[1])
+      end
+    end
+    total
+  end
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -4,4 +4,12 @@ class FoodTruck
     @name = name
     @inventory = {}
   end
+
+  def check_stock(item)
+    if @inventory.any?{|stocked_item| item.name == stocked_item[:name]}
+      require "pry";binding.pry
+    else
+      return 0
+    end
+  end
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -12,4 +12,13 @@ class FoodTruck
       return 0
     end
   end
+
+  def stock(new_item, amount)
+    if @inventory[new_item] == nil
+      @inventory[new_item] = amount
+    else
+      @inventory[new_item] += amount
+    end
+  end
+  require "pry";binding.pry
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -6,8 +6,10 @@ class FoodTruck
   end
 
   def check_stock(item)
-    if @inventory.any?{|stocked_item| item.name == stocked_item[:name]}
-      require "pry";binding.pry
+    if @inventory.find do |stocked_item|
+        item == stocked_item[0]
+      end != nil
+      @inventory[item]
     else
       return 0
     end
@@ -20,5 +22,4 @@ class FoodTruck
       @inventory[new_item] += amount
     end
   end
-  require "pry";binding.pry
 end
--- a/lib/item.rb
+++ b/lib/item.rb
@@ -1,0 +1,7 @@
+class Item
+  attr_reader :name, :price
+  def initialize(info)
+    @name = info[:name]
+    @price = info[:price].gsub(/[^0-9a-z.]/i, '').to_f
+  end
+end
--- a/test/event_test.rb
+++ b/test/event_test.rb
@@ -36,4 +36,8 @@ class EventTest < Minitest::Test
   def test_food_truck_names
     assert_equal ["Rocky Mountain Pies", "Ba-Nom-a-Nom", "Palisade Peach Shack"] ,@event.food_truck_names
   end
+
+  def test_food_trucks_that_sell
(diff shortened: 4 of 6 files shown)


Gallup93:
Got bogged down on iteration one for like 15 mins too long. After that, I never felt like I didn't know what to do but I ran out of time.

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+    @name = name
+    @food_trucks = []
+  end
+
+  def add_food_truck(truck)
+    @food_trucks << truck
+  end
+
+  def food_truck_names
+    @food_trucks.map {|truck| truck.name}
```
aziobrow, line 13: great tool for this job

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |truck|
+      truck.check_stock(item) > 0
```
aziobrow, line 18: nice use of a helper method here

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+  def total_inventory
+    totals = {}
+    sorted_item_list.each do |item|
+      totals[item] = 0
+    end
+    @food_trucks.each do |truck|
+      truck.inventory.keys.each do |key|
+        totals[key.name] += truck.inventory[key]
+      end
+    end
```
aziobrow, line 39: not quite. you have a few pieces that don't work here.  

I'd do something like this:

```
  def total_inventory
    result = {}
        
    items.each do |item|
    end
    result
  end
```

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+      truck.check_stock(item) > 0
+    end
+  end
+
+  def sorted_item_list
+    list = []
+    @food_trucks.each do |truck|
+      truck.inventory.keys.each {|key| list << key.name}
+    end
+    list.uniq.sort
```
aziobrow, line 27: cool-- nice work with this one. it definitely does what you need for it to do. 

That's definitely the important part, but do note that you're iterating a ton of times here-- you're using nested `each`s, so you're iterating several times for every truck (let's say you have 4 trucks and each sells 3 items-- you're iterating 3 times for each of the 4 trucks, which is exponential) and then you're iterating again at the end with the `sort`.  This is fine with a small set of data as you have here, but if this were a set of data like you're working with in Futbol, this would be really inefficient (in the example above, a total of 5 separate iterations over 5 different collections-- once for the inventory of each of the 4 trucks and once more over the list of resulting items)

Something like this would really cut it down:

```
@food_trucks.flat_map {|truck| truck.inventory.keys}.uniq.sort
```
This is just two iterations (once over the trucks, and once over the resulting items).

Hopefully this made sense-- let me know if it doesn't and I can try to clarify tomorrow.

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = {}
```

`@inventory = Hash.new(0)`

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def check_stock(item)
+    if @inventory.find do |stocked_item|
+        item == stocked_item[0]
+      end != nil
+      @inventory[item]
+    else
+      return 0
+    end
```
aziobrow, line 15: nice! it's great that you caught that you had to check for whether or not the item existed in `@inventory`.

aziobrow: if you had the default value set, as I mentioned above, you don't actually have to check for the include (since it would be 0 if it didn't exist).

So this method could just be:
```
  def check_stock(item)
      @inventory[item]
 end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def stock(new_item, amount)
+    if @inventory[new_item] == nil
+      @inventory[new_item] = amount
+    else
+      @inventory[new_item] += amount
+    end
```
aziobrow, line 23: same thing here-- this is great for the way you have it written!

```
def stock(item, price)
    @inventory[item] += price
 end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def potential_revenue
+    total = 0
+    if inventory != nil
+      @inventory.each do |item|
```
aziobrow, line 29: this is a good implementation, although when iterating over a hash, you should split apart the block variables into the key and value-- it makes it much easier to read/understand. 

```
def potential_revenue
    total = 0
         item.price * quantity
     end
     total 
end 
```

aziobrow: see how that's much easier to figure out what's going on then trying to figure out what item[0].price/item[1] is.

lib/item.rb
```diff
@@ -0,0 +1,7 @@
+class Item
+  attr_reader :name, :price
+  def initialize(info)
+    @name = info[:name]
+    @price = info[:price].gsub(/[^0-9a-z.]/i, '').to_f
```
aziobrow, line 5: cool. it would have been fine to explicitly check for `$` as your first argument for `gsub` instead of using regex here, but what you have here is great too.

test/event_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    @item4 = Item.new({name: "Banana Nice Cream", price: "$4.25"})
+    @item3 = Item.new({name: "Peach-Raspberry Nice Cream", price: "$5.30"})
+    @item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+    @food_truck1.stock(@item1, 35)
+    @food_truck1.stock(@item2, 7)
+    @food_truck2.stock(@item4, 50)
+    @food_truck2.stock(@item3, 25)
+    @food_truck3.stock(@item1, 65)
+    @event = Event.new("South Pearl Street Farmers Market")
+    @event.add_food_truck(@food_truck1)
```
aziobrow, line 22: there are workarounds to this, but you adding food trucks in your setup means that you can't test that the food_truck array starts out empty as its default value, which is a test you should have.

test/event_test.rb
```diff
@@ -0,0 +1,55 @@
...
+  def test_food_truck_names
+    assert_equal ["Rocky Mountain Pies", "Ba-Nom-a-Nom", "Palisade Peach Shack"] ,@event.food_truck_names
+  end
+
+  def test_food_trucks_that_sell
+    assert_equal [@food_truck1, @food_truck3], @event.food_trucks_that_sell(@item1)
+  end
+
+  def test_access_to_trucks_potential
+    assert_equal 345.00, @food_truck2.potential_revenue
```
aziobrow, line 45: you should check multiple revenues here to ensure that it still does the right thing (for example, a common issue is that floats don't handle the right number of decimal places, so checking a revenue that isn't an even dollar would help ensure that you didn't fall into that trap)

test/event_test.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def test_sorted_item_list
+    assert_equal ["Apple Pie (Slice)", "Banana Nice Cream", "Peach Pie (Slice)", "Peach-Raspberry Nice Cream"], @event.sorted_item_list
+  end
+
+  def test_total_inventory
+    assert_equal ({}) ,@event.total_inventory
```
aziobrow, line 53: this isn't a sufficient test. you wouldn't have gotten the points for this method if you had gotten it working, because this method is essentially untested, based on what you have here.  make sure to complete your tests before writing implementation.

test/food_truck_test.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def test_initialization_with_attributes
+    assert_instance_of FoodTruck, @food_truck
+    assert_equal "Rocky Mountain Pies", @food_truck.name
+    assert_equal ({}), @food_truck.inventory
+  end
+
+  def test_check_stock
+    assert_equal 0, @food_truck.check_stock(@item1)
```
aziobrow, line 19: it's great to check that an empty inventory returns 0 as the quantity for an item that hasn't been stocked yet, but i would definitely also check an item after it's been added to ensure that the correct quantities are being retrieved.

test/food_truck_test.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def test_stock_method
+    @food_truck.stock(@item1, 30)
+    assert_equal ({@item1 => 30}), @food_truck.inventory
+    assert_equal 30, @food_truck.check_stock(@item1)
```
aziobrow, line 25: since they are two separate methods, i'd split the test/assertions for check_stock from the inventory test/assertions.

test/food_truck_test.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def test_potential_revenue
+    @food_truck.stock(@item1, 30)
+    assert_equal (30 * 3.75), @food_truck.potential_revenue
```
aziobrow, line 30: you should check multiple revenues here to ensure that it still does the right thing (for example, a common issue is that floats don't handle the right number of decimal places, so checking a revenue that isn't an even dollar would help ensure that you didn't fall into that trap)

aziobrow: it would be preferable to calculate the expected answer-- tests should be as straightforward as possible and not require a reader to do any math/additional work to figure out what the expected output should be.

test/item_test.rb
```diff
@@ -0,0 +1,16 @@
...
+  def setup
+    @item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    @item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+  end
+
+  def test_initialization_with_attributes
+    assert_instance_of Item, @item1
+    assert_equal "Apple Pie (Slice)", @item2.name
+    assert_equal 2.50, @item2.price
+  end
```
aziobrow, line 15: 👍

lib/food_truck.rb
aziobrow: i thiiink i learned that you can do `sum` on a hash (you didn't used to be able to so I think i've just been out of touch)

```
def potential_revenue
         item.price * quantity
     end
end 
```

test/item_test.rb
Gallup93: Appreciate the work you put into the feedback. Thank you!

PR #19, opened 2020-04-15
Eric Larson

EricLarson2020:


timomitchel:
Would have loved to see more commits along the way and those tests committed first! This is for your own habit building benefit!

lib/item.rb
```diff
@@ -0,0 +1,13 @@
+class Item
+  attr_reader :name
+  def initialize(info)
+    @name = info[:name]
+    @price = info[:price]
+  end
+
+
+  def price
+
```
timomitchel, line 10: Spacing can be cleaned up here, we always want indentation and spacing within our methods to be consistent so that our eyes can more easily recognize the correct and incorrect patterns as we become more familiar with the code. It's preferred to do this conversion right in initialize and then used and attr_reader for `price`: 
`@price =info[:price].gsub("$", "").to_f`
This way the conversion only runs once, not every time you call the `#price` method.

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
+class FoodTruck
+  attr_reader :name
+  attr_accessor :inventory
```
timomitchel, line 3: No need for an accessor here. We don't need to change this value outside of a `FoodTruck` instance, so we only need a reader.

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize(name)
+    @name = name
+    @inventory = {}
```
timomitchel, line 6: We were hoping to see this hash initialized with a default value of 0 based on the interaction patter. You can do this two ways:
```
Hash.new(0)
```
```
@inventory.default = 0
```

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def check_stock(item)
+    if @inventory[item] == nil
```
timomitchel, line 10: This method needed to return the value for the item key in our inventory hash:
`@inventory[item]`

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
...
+      return 0
+    end
+    # require"pry";binding.pry
+    @inventory[item]
+  end
+
+  def stock(item, quanitity)
+    if @inventory[item] == nil
```
timomitchel, line 18: This is solid control flow to get the job done. `@inventory[item].nil?` reads a little bit better.

lib/food_truck.rb
```diff
@@ -0,0 +1,35 @@
...
+    @inventory[item] = quanitity
+    else
+      @inventory[item] += quanitity
+    end
+  end
+
+  def potential_revenue
```
timomitchel, line 25: Great job with this method. This is how we want to see you iterate over hashes. The `sum` enumerable can also take a block to save you a few lines:
```
inventory.sum do |key, value|
      key.price * value
end
```

lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+    @name = name
+    @food_trucks = []
+  end
+
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    @food_trucks.map do |truck|
```
timomitchel, line 13: The perfect enumerable for the job here.

lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+    item_name = item.name
+
+    trucks_sell = @food_trucks.select do |truck|
+
+      truck.inventory.keys.each do |inventory|
+          inventory.name == item_name
+
+        end
+    end
+    trucks_sell
```
timomitchel, line 29: Not quite. This method was best suited for a `find_all` combined with an `include?`, which is a common pattern in ruby:
`@food_trucks.find_all {|truck| truck.inventory.include?(item)}`
We know we want a subset of FoodTruck objects as our return value based on the condition of whether that FoodTruck object has the item within its inventory hash.

lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+       if complete_inventory[item] == nil
+         complete_inventory[item] = {
+
+           quantity: total_quantity(item),
+           food_trucks: food_trucks_that_sell(item)
+         }
+    end
+  end
+end
+complete_inventory
```
timomitchel, line 49: It's hard to tell if this value is inside or outside of your method scope. This is why indentation and spacing are important, specifically when our methods become this complex.

lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+      end
+    end
+    over_items
+  end
+
+  def total_quantity(current_item)
+    quantity = 0
+    @food_trucks.each do |truck|
+
+      truck.inventory.each do |item|
```
timomitchel, line 72: This block value should be split into a `|key, value|` pair.

lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+
+
+end
+
+  def overstocked_items
+    over_items = []
+    @food_trucks.each do |truck|
+
+      truck.inventory.keys.each do |item|
+        if total_quantity(item) > 50
```
timomitchel, line 60: This is close, but we need one more condition here and that is the item being sold in more than one food_truck, which would be an easier task if we started iterating on the return value of total_inventory and then had access to those number for each food_truck. This is why it's always best to make sure you have a method working before moving on to the next in the interaction pattern.

test/event_test.rb
```diff
@@ -0,0 +1,158 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/item'
+require './lib/food_truck'
+require './lib/event'
+
+class EventTest < Minitest::Test
```
timomitchel, line 7: This test would benefit greatly from a `setup` method. Saving you precious time you're spending copying and pasting for object creation.

test/event_test.rb
```diff
@@ -0,0 +1,158 @@
...
+    food_truck2.stock(item4, 50)
+    food_truck2.stock(item3, 25)
+    food_truck3 = FoodTruck.new("Palisade Peach Shack")
+    food_truck3.stock(item1, 65)
+    event.add_food_truck(food_truck1)
+    event.add_food_truck(food_truck2)
+    event.add_food_truck(food_truck3)
+    assert_equal 148.75, food_truck1.potential_revenue
+    assert_equal 345.00, food_truck2.potential_revenue
+    assert_equal 243.75, food_truck3.potential_revenue
```
timomitchel, line 78: Great job with this test setup and execution.

test/event_test.rb
```diff
@@ -0,0 +1,158 @@
...
+    food_truck1.stock(item2, 7)
+    food_truck2 = FoodTruck.new("Ba-Nom-a-Nom")
+    food_truck2.stock(item4, 50)
+    food_truck2.stock(item3, 25)
+    food_truck3 = FoodTruck.new("Palisade Peach Shack")
+    food_truck3.stock(item1, 65)
+    event.add_food_truck(food_truck1)
+    event.add_food_truck(food_truck2)
+    event.add_food_truck(food_truck3)
+    # assert_equal 0, event.total_inventory
```
timomitchel, line 98: Yea, I would not have moved on from here. Following TDD really matters when the interaction becomes more complex.

test/event_test.rb
```diff
@@ -0,0 +1,158 @@
...
+    food_truck1.stock(item2, 7)
+    food_truck2 = FoodTruck.new("Ba-Nom-a-Nom")
+    food_truck2.stock(item4, 50)
+    food_truck2.stock(item3, 25)
+    food_truck3 = FoodTruck.new("Palisade Peach Shack")
+    food_truck3.stock(item1, 65)
+    event.add_food_truck(food_truck1)
+    event.add_food_truck(food_truck2)
+    event.add_food_truck(food_truck3)
+    assert_equal [item1], event.overstocked_items
```
timomitchel, line 119: This is the correct test setup here.

test/food_truck_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+    assert_equal "Apple Pie (Slice)", item2.name
+    assert_equal 2.50, item2.price
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+    assert_instance_of FoodTruck, food_truck
+  end
+
+  def test_it_has_attributes
+    item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
```
timomitchel, line 19: No need for the `Item` instances, because we don't need them in order to test the `FoodTruck` attributes.

test/food_truck_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    assert_equal ({item1 => 30}),food_truck.inventory
+    assert_equal 30, food_truck.check_stock(item1)
+    food_truck.stock(item1, 25)
+    assert_equal 55, food_truck.check_stock(item1)
+    food_truck.stock(item2, 12)
+    expected = {
+      item1 => 55,
+      item2 => 12
+    }
+    assert_equal expected, food_truck.inventory
```
timomitchel, line 40: Great test here.

PR #20, opened 2020-04-15
Madeleine Halle

madhalle:


lib/event.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def add_food_truck(food_truck_object)
+    @food_trucks << food_truck_object
+  end
+
+  def food_truck_names
+    names = @food_trucks.map do |foodtruck|
+      foodtruck.name
+    end
+    names
```
aziobrow, line 16: great use of map! perfect tool here. remember, since you're not using `names` for anything else, the `map` itself can be the last line of this code and still work correctly. 

```
  def food_truck_names
    @food_trucks.map do |foodtruck|
       foodtruck.name
    end
end 
```

lib/event.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |foodtruck|
+      foodtruck.inventory.include?(item)
```
aziobrow, line 21: beautiful. also perfect tool here.

lib/food_truck.rb
```diff
@@ -0,0 +1,30 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = {}
```

`@inventory = Hash.new(0)`

lib/food_truck.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def check_stock(item)
+    if inventory.include?(item)
+      @inventory[item]
+    else
+      0
+    end
```
aziobrow, line 13: nice! it's great that you caught that you had to check for whether or not the item existed in @inventory.

aziobrow: if you had the default value set, as I mentioned above, you don't actually have to check for the include (since it would be 0 if it didn't exist).

So this method could just be:
```
  def check_stock(item)
      @inventory[item]
 end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def stock(item, quantity)
+    #@inventory[item] = quantity
+    if @inventory[item] == nil
+      @inventory[item] = 0
+    end
+    @inventory[item] += quantity
```
aziobrow, line 21: same thing here-- this is great for the way you have it written!

```
def stock(item, price)
    @inventory[item] += price
 end 
```

lib/food_truck.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def potential_revenue
+    @inventory.keys.sum do |key|
+      (key.price[1..-1].to_f) * @inventory[key]
```
aziobrow, line 26: great job getting this to work-- there were some tricky pieces here, and this does exactly what you need it to do.  a couple things:

1) since you need both the key (the item) and the value (its quantity) for each inventory element, you could have just iterated over the hash itself and used two block variables, which would make this method much more readable

2) the stuff you had to do to adjust the price-- that should have been done in `initialize` for the item (you missed that), so you wouldn't have had to do that here.  

So something like this: 

``` 
def potential_revenue
    total = 0
      item.price * quantity
     end
end 
```
Notice that things are much easier to read when they are descriptive, rather than `key`

lib/item.rb
```diff
@@ -0,0 +1,7 @@
+class Item
+  attr_reader :name, :price
+  def initialize(item_info)
+    @name = item_info[:name]
+    @price = item_info[:price]
```
aziobrow, line 5: i mentioned this above, but you missed this. the interaction pattern asks you to change this price to a float here.

test/event_test.rb
```diff
@@ -0,0 +1,148 @@
...
+  end
+
+  def test_add_food_truck
+    @food_truck1.stock(@item1, 35)
+    @food_truck1.stock(@item2, 7)
+
+    @food_truck2.stock(@item4, 50)
+    @food_truck2.stock(@item3, 25)
+
+    @food_truck3.stock(@item1, 65)
```
aziobrow, line 38: you don't need all this stocking setup for this method-- it doesn't hurt, but probably not a good use of your time.

test/event_test.rb
```diff
@@ -0,0 +1,148 @@
...
+  end
+
+  def test_food_truck_names
+    @food_truck1.stock(@item1, 35)
+    @food_truck1.stock(@item2, 7)
+
+    @food_truck2.stock(@item4, 50)
+    @food_truck2.stock(@item3, 25)
+
+    @food_truck3.stock(@item1, 65)
```
aziobrow, line 54: same here-- stocking is unnecessary to test this method

test/event_test.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+
+    assert_equal [@food_truck1, @food_truck2, @food_truck3] ,@event.food_trucks
```
aziobrow, line 60: this assertion on line 60 shouldn't be here-- it's a different method that's already been tested above.

test/food_truck_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    assert_equal 0, @food_truck.check_stock(@item1)
+    @food_truck.stock(@item1, 30)
+    expected = {@item1 => 30}
+    assert_equal expected, @food_truck.inventory
+    assert_equal 30, @food_truck.check_stock(@item1)
+    @food_truck.stock(@item1, 25)
+    assert_equal 55, @food_truck.check_stock(@item1)
+    @food_truck.stock(@item2, 12)
+    expected = {@item1 => 55, @item2 => 12}
+    assert_equal expected, @food_truck.inventory
```
aziobrow, line 40: i would split these out-- i would do a test for just the `check_stock` assertions, and a different test for the `inventory` assertions.

test/item_test.rb
```diff
@@ -0,0 +1,16 @@
...
+class ItemTest < Minitest::Test
+  def setup
+    @item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    @item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+  end
+
+  def test_it_exists
+    assert_instance_of Item, @item1
+    assert_instance_of Item, @item2
+  end
```
aziobrow, line 15: you're missing your attributes test here, which is maybe why you didn't catch the mistake you made with the price

lib/event.rb
```diff
@@ -0,0 +1,44 @@
...
+          inner_hash[:food_trucks] = []
+          new_hash[item] = inner_hash
+        elsif new_hash.include?(item)
+          inner_hash[:quantity] += item_quantity
+          inner_hash[:food_trucks] << foodtruck
+          new_hash[item] = inner_hash
+        end
+      end
+    end
+    new_hash
```
aziobrow, line 42: this one was super tricky, and you showed really good effort-- this looks like some super solid productive struggle! your quantities are a little off here-- i'd probably do something like this to make everything super clear:

```
  def total_inventory
    result = {}
        
    items.each do |item|
    end
    result
  end
```

lib/food_truck.rb
aziobrow: i thiiink i learned that you can do sum on a hash (you didn't used to be able to so I think i've just been out of touch)
```
def potential_revenue
         item.price * quantity
     end
end 
```

PR #21, opened 2020-04-15
Taylor Keller

takeller:
Was sooooo close to getting the sell method to work. I think I would have had it with a few more minutes.

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -5,4 +5,8 @@ class Event
     @name = name
     @food_trucks = Array.new
   end
+
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -48,6 +48,12 @@ class Event
     total_inventory.to_h
   end
 
+  def sorted_item_list
+    item_list = items_in_event.keys
+    named_item_list = item_list.map { |item| item.name  }
+    sorted_list = named_item_list.sort
+  end
+
 
 end
 
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,7 @@
+class Event
+
+  attr_reader :name 
+  def initialize(name)
+    @name = name
+  end
+end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -16,7 +16,7 @@ class Event
 
   def food_trucks_that_sell(item)
     food_trucks_with_item = @food_trucks.find_all do |food_truck|
-      food_truck.inventory.has_key?(item)
-    end 
+      food_truck.check_stock(item) > 0  
+    end
   end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -11,6 +11,12 @@ class Event
   end
 
   def food_truck_names
-    food_trucks = @food_trucks.map { |food_truck| food_truck.name } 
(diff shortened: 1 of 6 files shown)


lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+    food_trucks = @food_trucks.map { |food_truck| food_truck.name }
+  end
+
+  def food_trucks_that_sell(item)
+    food_trucks_with_item = @food_trucks.find_all do |food_truck|
+      food_truck.check_stock(item) > 0
+    end
+  end
+
+  def items_in_event
```
memcmahon, line 24: Love these helper methods!

lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def total_inventory
+    total_inventory = items_in_event
+    total_inventory.map do |item, item_info|
+      item_info[:quantity] = total_quantity(item)
+      item_info[:food_trucks] = food_trucks_that_sell(item)
+    end
+    total_inventory.to_h
+  end
```
memcmahon, line 49: Great job with this method - though you do not _need_ to assign a variable for `total_inventory`.  You could consolidate like this:

```ruby
items_in_event.map do |item, item_info|
      item_info[:quantity] = total_quantity(item)
      item_info[:food_trucks] = food_trucks_that_sell(item)
end.to_h
```

lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def sorted_item_list
+    item_list = items_in_event.keys
```
memcmahon, line 52: Again, you don't need to assign this variable, you could call `.map` directly on `item_in_event.keys`

lib/event.rb
```diff
@@ -0,0 +1,86 @@
...
+        food_truck.inventory[item] - quantity_remaining
+      else 
+        quantity_remaining = quantity - food_truck.inventory[item]
+        food_truck.inventory[item] = 0
+      end
+
+    end
+    true
+
+  end
```
memcmahon, line 84: You are sooooo close! Definitely on the right track - keep in mind that the resulting quantity for the inventory will not always be 0 but will always be reduced.

test/event_test.rb
```diff
@@ -0,0 +1,149 @@
...
+    assert_equal expected, @event.sorted_item_list
+  end
+
+  def test_overstocked_items
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal [@item1], @event.overstocked_items
+  end
+
```
memcmahon, line 132: Looks like you're missing the `date` method here!

test/item_test.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def test_it_exists
+    assert_instance_of Item, @item1
+    assert_instance_of Item, @item2
+  end
+
+  def test_it_has_readable_attributes
+    assert_equal 'Peach Pie (Slice)', @item1.name
+    assert_equal 'Apple Pie (Slice)', @item2.name
+    assert_equal "$3.75", @item1.price
```
memcmahon, line 20: Keep a close eye on the interaction pattern - `price` should return a float, not a string!

lib/event.rb
takeller: This makes sense. I haven't become super comfortable with tagging methods on an end like this, but I'll give it a try.

PR #22, opened 2020-04-15
James Belta

JBelta:


lib/food_truck.rb
```diff
@@ -0,0 +1,24 @@
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new
```
timomitchel, line 6: We were hoping to see this hash initialized with a default value of 0 based on the interaction patter. You can do this two ways:
```
Hash.new(0)
```
```
@inventory.default = 0
```

lib/food_truck.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def check_stock(item)
+    if @inventory == {}
```
timomitchel, line 10: This method needed to return the value for the item key in our inventory hash:
`@inventory[item]`

lib/item.rb
```diff
@@ -0,0 +1,9 @@
+class Item
+  attr_reader :name, :price
+
+  def initialize(info)
+    @name = info[:name]
+    @price = info[:price]
```
timomitchel, line 6: We needed to convert this `@price` string into a float here and get rid of the dollar sign that was prepended to the string. 
`info[:price].delete('$').to_f`

lib/food_truck.rb
```diff
@@ -0,0 +1,24 @@
...
+      @inventory.count
+    else
+     @inventory[item]
+   end
+  end
+
+  def stock(item, quantity)
```
timomitchel, line 17: Nice job with this method! This works! However, since `has_key?` will always return a boolean, we can simplify the logic of the conditional a bit:
```
if @inventory.has_key?(item)
@inventory[item] += quantity
else
@inventory[item] = quantity
end
```
When the condition returns true, we add to the current value of item, when it's false, we add the item key to the hash and assign it a quantity.

test/food_truck_test.rb
```diff
@@ -0,0 +1,62 @@
...
+    food_truck = FoodTruck.new("Rocky Mountain Pies")
+
+    food_truck.stock(item1, 30)
+    assert_equal 30, food_truck.check_stock(item1)
+
+    food_truck.stock(item1, 25)
+    assert_equal 55, food_truck.check_stock(item1)
+  end
+
+  def test_total_truck_stock
```
timomitchel, line 48: All of the tests before this one are excellent. But this test does not follow the interaction pattern. Remember that `inventory` holds a hash of items and quantities within a `food_truck` instance, not an array of items. Line 60 never had a chance unfortunately.

test/item_test.rb
```diff
@@ -0,0 +1,22 @@
...
+  def test_it_exist
+    item = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+
+    assert_instance_of Item, item
+  end
+
+  def test_it_has_attributes
+    item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+
+    assert_equal "Apple Pie (Slice)", item2.name
```
timomitchel, line 16: Remember to follow the interaction pattern exactly, so that you don't lose points when you need to convert a data type.

PR #23, opened 2020-04-15
Final commit, Eric Hale

EHale64:
learned what the merge and merge! methods do today...

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,6 @@
+class Event
+
+  def initialize(argument)
+    @argument = argument
+  end
+end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,6 +1,34 @@
 class Event
+  attr_reader :name, :food_trucks
 
-  def initialize(argument)
-    @argument = argument
+  def initialize(name)
+    @name = name
+    @food_trucks = []
+  end
+
+  def add_food_truck(truck)
+     @food_trucks << truck
+  end
+
+  def food_truck_names
+    @food_trucks.map do |truck|
+      truck.name
+    end
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |truck|
+      truck if truck.inventory.include?(item)
+    end
+  end
+
+  def total_inventory
+    total_inventory = Hash.new
+    @food_trucks.each do |truck|
+      truck.inventory.each do |item|
+        total_inventory[item] = {"quantity": nil, "food_trucks": []}
+      end
+    end
+    require 'pry': binding.pry
   end
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -12,9 +12,12 @@ class FoodTruck
   end
 
   def stock(item, item_quantity)
-    stock = Hash.new
-    stock[item] = item_quantity
-    old_stock = @inventory
-    old_stock.merge!(stock) {|key, old_value, new_value| old_value + new_value}
+    @inventory[item] += item_quantity
+  end
+
+  def potential_revenue
+     @inventory.map do |item, item_quantity|
+       @inventory[item]  *  item.price
+     end.sum
   end
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -1,0 +1,20 @@
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
+    @inventory.default= 0
+  end
+
+  def check_stock(item)
+    @inventory[item]
+  end
+
+  def stock(item, item_quantity)
+    stock = Hash.new
+    stock[item] = item_quantity
+    old_stock = @inventory
+    old_stock.merge!(stock) {|key, old_value, new_value| old_value + new_value}
+  end
+end
--- a/lib/item.rb
+++ b/lib/item.rb
@@ -1,0 +1,13 @@
+class Item
+  attr_reader :name
+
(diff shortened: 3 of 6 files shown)


lib/event.rb
```diff
@@ -0,0 +1,6 @@
+class Event
+
+  def initialize(argument)
+    @argument = argument
```
aziobrow, line 4: nice job seeing it's only one argument-- however, not a best practice to call an argument `argument` (technically, it's a parameter, not an argument). more to the point, you could call any argument `argument` for any method, so be precise about what it is.  this is the event's `name`.

lib/food_truck.rb
```diff
@@ -0,0 +1,20 @@
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = {}
+    @inventory.default= 0
```
aziobrow, line 7: you could also say `@inventory = Hash.new(0)` to accomplish this in one line.

``` 
 def initialize(name)
    @name = name
    @inventory = Hash.new(0)
end
```

Nice job giving it a default value either way! This is a good use case.

lib/event.rb
aziobrow: Oh... in hindsight, this may be where you ran out of time and maybe it auto-filled for you. 👍

lib/food_truck.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def check_stock(item)
+    @inventory[item]
```
aziobrow, line 11: great!

lib/food_truck.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def stock(item, item_quantity)
+    stock = Hash.new
+    stock[item] = item_quantity
+    old_stock = @inventory
+    old_stock.merge!(stock) {|key, old_value, new_value| old_value + new_value}
```
aziobrow, line 18: this works, and you must have spent some good productive struggle researching this and getting it to do what you wanted it to do! nice work .

However, with a fundamental understanding of how hashes work, this could have been done much more easily:

```
def stock(item, item_quantity)
    @inventory[item] += item_quantity
 end 
```

Since items are keys in the inventory, you can just use the item being passed in to access the quantity of that item in the inventory hash. if the item doesn't already exist, since you used a default of 0, it will add that key/value pair.  if it already existed, it will simply find that item and add to its quantity.

lib/item.rb
```diff
@@ -0,0 +1,13 @@
+class Item
+  attr_reader :name
+
+  def initialize(item_info)
+    @name = item_info[:name]
+    @price = item_info[:price]
```
aziobrow, line 6: nice handling of the hash here.

lib/item.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def price
+  number = @price.delete"$"
+  number.to_f
```
aziobrow, line 11: also nice job writing out this method and catching what you needed to do to turn this into a float as the interaction pattern asked.

lib/event.rb
aziobrow: oh-- in hindsight, this looks like where you ran out of time, so i imagine that this autocompleted for you 👍

test/food_truck_test.rb
```diff
@@ -0,0 +1,37 @@
...
+    assert_instance_of FoodTruck, @food_truck
+  end
+
+  def test_it_has_attributes
+    assert_equal "Rocky Mountain Pies", @food_truck.name
+    assert_equal ({}), @food_truck.inventory
+  end
+
+  def test_it_can_return_an_item_in_stock
+    assert_equal 0, @food_truck.check_stock(@item1)
```
aziobrow, line 24: it's great to check that an empty inventory returns 0 as the quantity for an item that hasn't been stocked yet, but i would definitely also check an item after it's been added to ensure that the correct quantities are being retrieved.

test/food_truck_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def test_it_can_stock_an_item
+      @food_truck.stock(@item1, 30)
+    assert_equal ({@item1 => 30}), @food_truck.inventory
+    assert_equal 30, @food_truck.check_stock(@item1)
```
aziobrow, line 30: oh i see-- you did it here. so i'd move these `check_stock` assertions into the above test, since you should really only be testing one method per test, as a general rule.

test/food_truck_test.rb
```diff
@@ -0,0 +1,37 @@
...
+      @food_truck.stock(@item1, 25)
+    assert_equal 55, @food_truck.check_stock(@item1)
+      @food_truck.stock(@item2, 12)
+      bindng.pry
+    assert_equal ({@item1 => 55, @item2 => 12}), @food_truck.inventory
```
aziobrow, line 35: nice

test/item_test.rb
```diff
@@ -0,0 +1,23 @@
...
+  def test_it_exists
+    assert_instance_of Item, @item1
+    assert_instance_of Item, @item2
+  end
+
+  def test_it_has_attributes
+    assert_equal 'Peach Pie (Slice)', @item1.name
+    assert_equal 3.75, @item1.price
+    assert_equal 'Apple Pie (Slice)', @item2.name
+    assert_equal 2.50, @item2.price
```
aziobrow, line 21: although more testing is better than less, you really only need to do one item here-- unless item1 was fundamentally different from item2 in some way, there's no reason to test multiples.

PR #24, opened 2020-04-15
Ashkan Abbasi

Ashkanthegreat:
Was working on last method in It 3 when I ran out of time.

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,20 @@
+class Event
+  attr_reader :name, :food_trucks
+
+  def initialize(name)
+    @name = name
+    @food_trucks = []
+  end
+
+  def add_food_truck(food_truck)
+    food_trucks << food_truck
+  end
+
+  def food_truck_names
+    @food_trucks.map { |food_truck| food_truck.name }
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.select { |food_truck| food_truck.check_stock(item) > 0 }
+  end
+end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -41,10 +41,11 @@ class Event
   end
 
   def sorted_item_list
-
-    total_inventory.find_all do |item, food_truck|
-      item.name 
+    sorted_items = []
+    total_inventory.each do |item, food_truck|
+      sorted_items << item.name
     end
+    sorted_items.sort
   end
 
 
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -17,4 +17,35 @@ class Event
   def food_trucks_that_sell(item)
     @food_trucks.select { |food_truck| food_truck.check_stock(item) > 0 }
   end
+
+  def total_inventory
+    result = {}
+    @food_trucks.each do |food_truck|
+      food_truck.inventory.each do |item, quantity|
+        result[item] = {quantity: 0, food_trucks: []} if result[item].nil?
+        result[item][:quantity] += quantity
+        result[item][:food_trucks] << food_truck
+      end
+    end
+    result
+  end
+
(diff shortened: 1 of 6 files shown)


lib/event.rb
```diff
@@ -0,0 +1,52 @@
...
+    @name = name
+    @food_trucks = []
+  end
+
+  def add_food_truck(food_truck)
+    food_trucks << food_truck
+  end
+
+  def food_truck_names
+    @food_trucks.map { |food_truck| food_truck.name }
```
aziobrow, line 14: good use of map

lib/event.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.select { |food_truck| food_truck.check_stock(item) > 0 }
```
aziobrow, line 18: nice select here

lib/event.rb
```diff
@@ -0,0 +1,52 @@
...
+  def total_inventory
+    result = {}
+    @food_trucks.each do |food_truck|
+      food_truck.inventory.each do |item, quantity|
+        result[item] = {quantity: 0, food_trucks: []} if result[item].nil?
+        result[item][:quantity] += quantity
+        result[item][:food_trucks] << food_truck
+      end
+    end
+    result
```
aziobrow, line 30: really nice job with this one. clean and easy to read.

lib/event.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def overstocked_items
+    overstocked_item = []
+     total_inventory.each do |item, food_truck|
+       if food_truck[:quantity] > 50 && food_truck[:food_trucks].length > 1
+         overstocked_item << item
+       end
+     end
+    overstocked_item
```
aziobrow, line 40: nice job here. a map would be the best tool, but this solution is very good.

lib/event.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def sorted_item_list
+    sorted_items = []
+    total_inventory.each do |item, food_truck|
```
aziobrow, line 45: again, this could be a map, but this works well.

lib/food_truck.rb
```diff
@@ -0,0 +1,22 @@
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
aziobrow, line 6: good use of a default value-- makes life a lot easier.

lib/food_truck.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def check_stock(item)
+    @inventory[item]
```
aziobrow, line 10: great

lib/food_truck.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def stock(item, quantity)
+    @inventory[item] += quantity
```
aziobrow, line 14: looks good

lib/food_truck.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def potential_revenue
+    @inventory.sum {|item, quantity|
```
aziobrow, line 18: good use of sum and good job splitting the two block variables, which makes for much better readability

lib/item.rb
```diff
@@ -0,0 +1,9 @@
+class Item
+  attr_reader :name, :price
+  
+  def initialize(item_params)
+    @name = item_params[:name]
+    @price = item_params[:price].delete('$').to_f
```
aziobrow, line 6: 👍

PR #25, opened 2020-04-15
Max Mitrani

Lithnotep:
Off my game a bit, and had trouble remembering to commit frequently.
Ran in to a touch of trouble on getting the price to a string. should have probably created another method for it. 
I will go back over this assessment when I have my head on straight.
Thank you!

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -22,4 +22,5 @@ class Event
     @food_trucks.find_all  {|truck| truck.inventory[item] > 0}
   end
 
+
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,25 @@
+require './lib/food_truck.rb'
+require './lib/item.rb'
+require 'pry'
+
+class Event
+  attr_reader :name, :food_trucks
+
+  def initialize(name)
+    @name = name
+    @food_trucks = []
+  end
+
+  def add_food_truck(truck)
+    @food_trucks.push(truck)
+  end
+
+  def food_truck_names
+    @food_trucks.map {|truck| truck.name}
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all  {|truck| truck.inventory[item] > 0}
+  end
+
+end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -1,0 +1,26 @@
+require './lib/item.rb'
+require 'pry'
+
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
+  end
+
+  def check_stock(item)
+    if @inventory[item] == nil
+      @inventory[item] = 0
+    else
+      @inventory[item]
+    end
+  end
+
+  def stock(item, amount)
+    check_stock(item)
+    @inventory[item] += amount
+  end
+
+
+end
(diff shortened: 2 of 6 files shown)


Lithnotep:
Just looked at my files and noticed my food truck and event wasn't saved when I commited. So it should just be spacing but if something critical is missing please let me know.

lib/event.rb
```diff
@@ -0,0 +1,26 @@
+require './lib/food_truck.rb'
+require './lib/item.rb'
```
memcmahon, line 2: You only need to require files when you use that class within the file that you're in - if you see `Item` or `FoodTruck` in the file.

lib/event.rb
```diff
@@ -0,0 +1,26 @@
...
+require 'pry'
+
+class Event
```
memcmahon, line 5: So far, this class looks great

lib/food_truck.rb
```diff
@@ -0,0 +1,36 @@
+require './lib/item.rb'
```
memcmahon, line 1: See my note in food_truck.rb about requiring files

lib/food_truck.rb
```diff
@@ -0,0 +1,36 @@
...
+require 'pry'
+
+class FoodTruck
+  attr_reader :name, :inventory
+
+  def initialize(name)
+    @name = name
+    @inventory = Hash.new(0)
```
memcmahon, line 9: Great use of a default value.

lib/food_truck.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def check_stock(item)
+    if @inventory[item] == nil
```
memcmahon, line 13: Since you have a default value for your `@inventory` hash, you do not need this conditional logic, you could just use `@inventory[item]`

lib/food_truck.rb
```diff
@@ -0,0 +1,36 @@
...
+      @inventory[item] = 0
+    else
+      @inventory[item]
+    end
+  end
+
+  def stock(item, amount)
+    check_stock(item)
```
memcmahon, line 21: You do not need to `check_stock` in this method, you only need the next line :)

lib/food_truck.rb
```diff
@@ -0,0 +1,36 @@
...
+    @inventory[item] += amount
+  end
+
+  def potential_revenue
+    revenue = 0
+    @inventory.each do |item|
+      item_value = (item[0].price).chop.to_f
```
memcmahon, line 28: On the right track - see my note in Item about the price.

lib/item.rb
```diff
@@ -0,0 +1,10 @@
+require 'pry'
+
+class Item
+  attr_reader :name, :price
+
+  def initialize(details)
+    @name = details[:name]
+    @price = details[:price]
```
memcmahon, line 8: Keep an eye on the interaction pattern - we wanted item.price to return a float (that would make the potential revenue a bit easier) A couple of ways to do this, but I think this is the most straightforward: `details[:price].delete('$').to_f`

test/item_test.rb
```diff
@@ -0,0 +1,24 @@
...
+    @item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
+  end
+
+  def test_it_exist
+    assert_instance_of Item, @item1
+  end
+
+  def test_has_attributes
+    assert_equal 'Peach Pie (Slice)' , @item1.name
+    assert_equal "$3.75", @item1.price
```
memcmahon, line 19: See my note on Item above, but next week make sure you are expecting the same thing as the interaction pattern!

PR #26, opened 2020-04-15
Kyle Iverson

kiverso:
Gets through most of iteration 3, but fails total_inventory_test.

### DIFF
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -12,4 +12,10 @@ class Event
   def food_truck_names
     @food_trucks.map { |truck| truck.name }
   end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |truck|
+      truck.inventory.keys.include?(item)
+    end
+  end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -4,4 +4,8 @@ class Event
     @name = name
     @food_trucks = food_trucks
   end
+
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -8,4 +8,8 @@ class Event
   def add_food_truck(food_truck)
     @food_trucks << food_truck
   end
+
+  def food_truck_names
+    @food_trucks.map { |truck| truck.name }
+  end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -25,6 +25,11 @@ class Event
     items.flatten.uniq
   end
 
+  def sorted_item_list
+  items = all_items.map { |item| item.name }
+  items.sort
+  end
+
   # def total_inventory
   #   @food_trucks.each do |truck|
   #
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,0 +1,6 @@
+class Event
+  def initialize(name)
+    @name = name
+    
+  end
+end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -18,4 +18,16 @@ class Event
       truck.inventory.keys.include?(item)
     end
   end
+
+  def all_items
+    items = @food_trucks.map { |truck| truck.inventory.keys }
+    # require "pry"; binding.pry
+    items.flatten.uniq
+  end
+
+  # def total_inventory
+  #   @food_trucks.each do |truck|
+  #
+  #   end
+  # end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -1,6 +1,7 @@
 class Event
-  def initialize(name)
+  attr_reader :name, :food_trucks
+  def initialize(name, food_trucks = [])
     @name = name
-    
+    @food_trucks = food_trucks
   end
 end
--- a/lib/event.rb
+++ b/lib/event.rb
@@ -30,9 +30,14 @@ class Event
   items.sort
   end
 
-  # def total_inventory
-  #   @food_trucks.each do |truck|
-  #
-  #   end
-  # end
+  def total_inventory
+    inventory = {}
+    all_items.each do |item|
+      unless inventory[item]
+        inventory[item] = {quantity: 0, food_trucks: food_trucks_that_sell(item) }
+      end
+      inventory[item][:quantity] = @food_trucks.sum{|truck| truck.check_stock(item)}
+    end
+    inventory
+  end
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -1,6 +1,7 @@
 class FoodTruck
-
-  def initialize(name)
-    
+  attr_reader :name, :inventory
+  def initialize(name, inventory = {})
+    @name = name
+    @inventory = inventory
   end
 end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -1,0 +1,6 @@
+class FoodTruck
+
+  def initialize(name)
+    
+  end
+end
--- a/lib/food_truck.rb
+++ b/lib/food_truck.rb
@@ -4,4 +4,16 @@ class FoodTruck
     @name = name
     @inventory = inventory
   end
+
+  def check_stock(item)
+    return 0 if !@inventory[item]
+    @inventory[item]
+  end
+
+  def stock(item, quantity)
+    unless @inventory[item]
+      @inventory[item] = 0
+    end
+    @inventory[item] += quantity
+  end
 end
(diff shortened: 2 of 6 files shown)


timomitchel:
Nice job committing throughout and showing us TDD!

lib/event.rb
```diff
@@ -0,0 +1,43 @@
+class Event
+  attr_reader :name, :food_trucks
+  def initialize(name, food_trucks = [])
```
timomitchel, line 3: This works, but generally we don't want to make the default argument unless there is an example of an argument being passed sometimes but not always. 
`@food_trucks = []` is preferred.

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    @food_trucks.map { |truck| truck.name }
+  end
+
+  def food_trucks_that_sell(item)
+    @food_trucks.find_all do |truck|
```
timomitchel, line 17: The perfect enumerable in conjunction with `.include?`, which also works directly on a hash. This would also have worked:
`truck.inventory.include?(item)`

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+      truck.inventory.keys.include?(item)
+    end
+  end
+
+  def all_items
+    items = @food_trucks.map { |truck| truck.inventory.keys }
```
timomitchel, line 23: I like that you broke this out into a helper method. Check out the enumerable `flat_map` which is more performant and does the flattening as you iterate.

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+    # require "pry"; binding.pry
+    items.flatten.uniq
+  end
+
+  def sorted_item_list
+  items = all_items.map { |item| item.name }
+  items.sort
```
timomitchel, line 30: Solid sorting method.

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def total_inventory
+    inventory = {}
+    all_items.each do |item|
+      unless inventory[item]
```
timomitchel, line 36: No need for this `unless` here, as we will set the key of an item for each item in the `all_items` array, so we don't need that check.

lib/event.rb
```diff
@@ -0,0 +1,43 @@
...
+        inventory[item] = {quantity: 0, food_trucks: food_trucks_that_sell(item) }
+      end
```
timomitchel, line 38: Great use of the return value of a previous method here.

lib/food_truck.rb
```diff
@@ -0,0 +1,28 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name, inventory = {})
+    @name = name
+    @inventory = inventory
```
timomitchel, line 5: Again, no need for the default value and extra argument. We wanted this hash initialized with a default value of 0:
`@inventory = Hash.new(0)`

lib/food_truck.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def check_stock(item)
+    return 0 if !@inventory[item]
```
timomitchel, line 9: Glad you got this to run, but with the inventory hash initialized with the default value of 0, this would simplify to:
`@inventory[item]`

lib/food_truck.rb
```diff
@@ -0,0 +1,28 @@
...
+    @inventory[item]
+  end
+
+  def stock(item, quantity)
+    unless @inventory[item]
+      @inventory[item] = 0
+    end
+    @inventory[item] += quantity
```
timomitchel, line 17: Similarly, this method would only need line 17:
`@inventory[item] += quantity`
if the default value of `@inventory` were 0.

lib/food_truck.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def potential_revenue
+    price_collection = @inventory.transform_keys do |item|
+      item.price
+    end
+    price_collection.sum do |price, quantity|
```
timomitchel, line 24: We could avoid iterating twice here: 
```
@inventory.sum do |item, amount|
   item.price * amount.to_f
end
```

test/event_test.rb
```diff
@@ -0,0 +1,92 @@
+require "minitest/autorun"
+require "./lib/food_truck"
+require "./lib/item"
+require "./lib/event"
+
+class EventTest < Minitest::Test
+  def setup
```
timomitchel, line 7: Nice use of the setup method to speed things up.

test/event_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    assert_equal expected, @event.name
+    assert_equal [], @event.food_trucks
+  end
+
+  def test_it_can_add_food_trucks
+    assert_equal [], @event.food_trucks
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    expected = [@food_truck1, @food_truck2, @food_truck3]
```
timomitchel, line 39: I love the use of an expected variable to clean things up.

test/event_test.rb
```diff
@@ -0,0 +1,92 @@
...
+  def test_it_can_get_food_truck_names
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    expected = ["Rocky Mountain Pies", "Ba-Nom-a-Nom", "Palisade Peach Shack"]
+    assert_equal expected, @event.food_truck_names
+  end
+
+  def test_it_can_get_food_trucks_that_sell_a_specific_item
+    @event.add_food_truck(@food_truck1)
```
timomitchel, line 52: Something to consider is adding a method that runs your `add_food_truck` setup. We don't want it in `setup`, because it would break our initial attributes test, but just like any other ruby class, we can create a helper method and then call it in every method that needs it: 
```
def add_trucks
@event.add_food_truck(@food_truck1)
@event.add_food_truck(@food_truck2)
@event.add_food_truck(@food_truck3)
end
```
Then we call `add_trucks` in each test method that needs that functionality.

test/event_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+
+    expected = {@item1 => {quantity: 100, food_trucks: [@food_truck1, @food_truck3]},
+                @item2 => {quantity: 7, food_trucks: [@food_truck2]},
+                @item3 => {quantity: 35, food_trucks: [@food_truck2, @food_truck3]},
+                @item4 => {quantity: 50, food_trucks: [@food_truck2]}
+                }
+    assert_equal expected, @event.total_inventory
```
timomitchel, line 90: Great job with this test setup.

test/event_test.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def test_it_can_get_food_trucks_that_sell_a_specific_item
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+    assert_equal [@food_truck1, @food_truck3], @event.food_trucks_that_sell(@item1)
+    assert_equal [@food_truck2], @event.food_trucks_that_sell(@item4)
+  end
+
+  def test_it_can_get_all_items
```
timomitchel, line 59: Great job testing your helper method.

test/event_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    @event.add_food_truck(@food_truck1)
+    @event.add_food_truck(@food_truck2)
+    @event.add_food_truck(@food_truck3)
+
+    assert_equal true, @event.all_items.include?(@item1)
```
timomitchel, line 64: You could have gone with an `assert_includes` here.

test/food_truck_test.rb
```diff
@@ -0,0 +1,48 @@
...
+    assert_equal 0, @food_truck.check_stock(@item1)
+    @food_truck.stock(@item1, 30)
+    @food_truck.stock(@item1, 25)
+    @food_truck.stock(@item2, 50)
+    assert_equal 55, @food_truck.check_stock(@item1)
+    expected = {@item1 => 55, @item2 => 50}
+    assert_equal expected, @food_truck.inventory
+  end
+  def test_it_can_get_potential_revenue
+    food_truck1 = FoodTruck.new("Rocky Mountain Pies")
```
timomitchel, line 32: Nice job setting this up with the local variables we needed from iteration 3.

PR #27, opened 2020-04-15
Jack Puchalla

JPuchalla:


lib/event.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def food_truck_names
+    @food_trucks.map { |truck| truck.name }
+  end
+
+  def food_trucks_that_sell(item)
+    #doesnt work
+    @food_trucks.find_all do |truck|
+      if !truck.inventory[item].nil?
```
memcmahon, line 20: Because you have a default value for your inventory hash, this will never return a nil value - it will be 0 or the inventory amount.

lib/food_truck.rb
```diff
@@ -0,0 +1,19 @@
...
+  def check_stock(item)
+  @inventory[item]
+  end
+
+  def stock(item, inventory)
+    @inventory[item] += inventory
+  end
+
+  def potential_revenue
+    @inventory.sum { |item, inventory| item.price * inventory}
```
memcmahon, line 17: Great use of sum on a hash!

lib/item.rb
```diff
@@ -0,0 +1,12 @@
+class Item
+  attr_reader :name
+  def initialize(item)
+    @name = item[:name]
+    @cost = item[:price]
+  end
+
+  def price
+    prc = @cost.delete! "$"
```
memcmahon, line 9: Good job with this, but if you are going to choose to not use parens, be consistent and don't use them anywhere else either.

PR #28, opened 2020-04-15
Judith Pillado 

judithpillado:
Got stuck on iteration 1 because I didn't copy and paste correctly (nervous) and that cost me a LOT of time. wasn't able to finish iteration 2 :(


lib/event.rb
```diff
@@ -0,0 +1,28 @@
...
+    @food_trucks = []
+  end
+
+  def add_food_truck(food_truck)
+    @food_trucks << food_truck
+  end
+
+  def food_truck_names
+    food_truck_name = []
+    food_truck_name << food_truck.name
```
aziobrow, line 15: you don't have anything accessible in this class called `food_truck`-- you only have `food_trucks`, which is a collection of individual food trucks. so to get their names, you'd need to iterate over each one and grab its name. 

```
  def food_truck_names
    food_truck_name = []
    @food_trucks.each do |food_truck|
         food_truck_name << food_truck.name
     end
     food_truck_names
end
```

lib/event.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  # def potential_revenue
+  #   @item.price.sum * amount
```
aziobrow, line 19: i see this is commented out, but you don't have anything called `@item` accessible in this class either, so you'd have to dig to get that too.

lib/food_truck.rb
```diff
@@ -0,0 +1,20 @@
+class FoodTruck
+  attr_reader :name, :inventory
+  def initialize(name)
+    @name = name
+    @inventory = {}
```

`@inventory = Hash.new(0)`

lib/food_truck.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def check_stock(amount)
+    amount = 0
```
aziobrow, line 9: not quite-- this method should take an item object, not an amount.  And you would check your inventory to see if you have that item, and if you do, you would want to use the item (as the key of the inventory hash) to access its value, which should be the amount of that item. 

Something like this:
```
  def check_stock(item)
    if @inventory[item] == nil
      @inventory[item] = 0
    else
      @inventory[item]
    end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,20 @@
...
+
```
aziobrow, line 10: if you had the default value set, as I mentioned above, you don't actually have to check for the nil value (since it would be 0 if it didn't exist).

So this method could just be:
```
  def check_stock(item)
      @inventory[item]
 end
```

lib/food_truck.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def stock(item, amount)
+    amount = Hash.new(0)
+    @inventory.each do |item, amount|
+      inventory[item] += amount
```
aziobrow, line 16: this would be really good for iterating over the inventory hash, so great job with the syntax for iterating over a hash... but we don't actually want to do that here. 

We want to stock a single, specific item in our already existing inventory with the specific amount being given to us.  So we just need to find that item in our inventory and then add the amount that's being passed in to its value. 

```
def stock(item, amount)
     if @inventory[item] == nil
        @inventory[item] = 0
      end
      @inventory[item] += price
end
```

aziobrow: Same here, if you chose to use the default value, this could just be
```
def stock(item, price)
    @inventory[item] += price
 end 
```

lib/item.rb
```diff
@@ -0,0 +1,8 @@
+class Item
+  attr_reader :name, :price
+
+  def initialize(attributes)
+    @name = attributes[:name]
+    @price = attributes[:price].to_s.tr('$', '').to_f
```
aziobrow, line 6: nice!

test/even_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+  #   @food_truck1.stock(@item1, 35)
+  #   @food_truck1.stock(@item2, 7)
+  #   @food_truck2.stock(@item4, 50)
+  #   @food_truck2.stock(@item3, 25)
+  #   @food_truck3.stock(@item1, 65)
+  #   @event.add_food_truck(@food_truck1)
+  #   @event.add_food_truck(@food_truck2)
+  #   @event.add_food_truck(@food_truck3)
+
```
aziobrow, line 50: these tests look great

test/food_truck_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    assert_instance_of FoodTruck, @food_truck
+  end
+
+  def test_it_has_attributes
+    assert_equal "Rocky Mountain Pies", @food_truck.name
+    assert_equal ({}), @food_truck.inventory
+  end
+
+  def test_it_can_check_stock
+    assert_equal 0, @food_truck.check_stock(@item1)
```
aziobrow, line 22: it's great to check that an empty inventory returns 0 as the quantity for an item that hasn't been stocked yet, but i would definitely also check an item after it's been added to ensure that the correct quantities are being retrieved.

test/food_truck_test.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def test_stock
+    @food_truck.stock(@item1, 30)
+    assert_equal ({@item1 => 30}), @food_truck.inventory
```
aziobrow, line 27: i would go farther on this one-- i'd check that you could add multiple different items, and also that you can add more of a single item.

test/food_truck_test.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def test_add_stock_and_inventory
+    @food_truck.stock(@item1, 30)
+    assert_equal 30, @food_truck.check_stock(@item1)
```
aziobrow, line 32: i'd put these assertions in the test above (your check_stock test), since that's the actual method you're testing here.

test/food_truck_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    @food_truck.stock(@item1, 25)
+    assert_equal 55, @food_truck.check_stock(@item1)
+    @food_truck.stock(@item2, 12)
+    assert_equal ({@item1 => 55, @item2 => 12}), @food_truck.inventory
```
aziobrow, line 36: and i'd put this in the test above, where you're checking stock (because `stock` is what updates `inventory`)

test/item_test.rb
```diff
@@ -0,0 +1,22 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/item'
+
+class ItemTest < Minitest::Test
+
+  def setup
+    @item1 = Item.new({name: 'Peach Pie (Slice)', price: "$3.75"})
+    @item2 = Item.new({name: 'Apple Pie (Slice)', price: '$2.50'})
```
aziobrow, line 9: you really only need one item here, although it never hurts to test more... just for the sake of time.

